PR #1, opened 2018-10-01
analyzer and example

alejoe91:
let's review code here. We can comment!


colehurwitz31:
Hello! A couple of things I noticed about the Analyzer class:

- inputExtractor --> getInputExtractor
- outputExtractor --> getOutputExtractor
- I think we should have gets for all the internal variable lists.
- I think all functions should return the thing it is computing maybe.. Like compute PCA returns pca components. Not sure about this one.

-unit_id as variable name --> should be unit_ids. I also think unit_ids should be a list only and not both. 
- Why do you have two return statements here:
https://github.com/alejoe91/spiketoolkit/blob/alessio/spiketoolkit/Analyzer.py#L192
I don't think I understand this part.

-cutout_start and cutout_end better than cutout list (or use tuple instead of list to store cutouts).
-- stop_frame -- > end_frame (to be the same as spikeinterface)
-- I would be a bit more verbose in some variable names like: i_w, fs, wf, etc. It makes easier for other people to look at the code later.

-- Having waveforms, waveform_list and self._waveforms might be a little confusing. Is there a need for all those different variables names.

-I think we should all agree on documentation so that the format of our documentation is the same across repos. 

-Looks good otherwise! The code is pretty readable which is nice!

alejoe91:
Hi Cole! Thanks for the comments.

- inputExtractor --> getInputExtractor --> OK
- I think we should have gets for all the internal variable lists. --> I would implement them as properties 
(not done it yet)
- I agree (PCA is not finished!)

- we can have unit_ids, but I think we should allow to have either an int, a list or None (to return and compute all) --> more flexile this way
- In case a single unit_ids is given, the function return an int (single element), otherwise it returns the list.

- cutout_start and cutout_end better than cutout list (or use tuple instead of list to store cutouts). --> mmm maybe cutout_start and cutout_end is better

- stop_frame -- > end_frame (to be the same as spikeinterface) --> OK
- I would be a bit more verbose in some variable names like: i_w, fs, wf, etc. It makes easier for other people to look at the code later. max_chan --> max_channel. --> OK

- Imports should be at the top of the file --> I placed sklearn import in the PCA functionas it's an heavy package. but if we need only PCA I can move it on top

- Having waveforms, waveform_list and self._waveforms might be a little confusing. Is there a need for all those different variables names. --> waveforms is the local variable to be returned and appended to waveform_list when multiple ids are provided. self._waveforms caches the computed waveforms so that they are not recomputed. I think we may need all 3

colehurwitz31:
Cool! I am glad we agree on that stuff :)

For the unit_ids, I think it should be a list in the documentation and people should try to use lists, but maybe we should check for an int and let the user get away with it? I don't really thing the input variable should be two types, but we can be flexible for the user. 

I think I understand the double return statements now. Although, I think it might still make sense to return a list with one element rather than that one element so that all the scripts are consistent when using our function. Not sure though, maybe Jeremy has an opinion about this.

PR #30, opened 2018-12-06
fix path error

espenhgn:


alejoe91:
Hi @espenhgn 

- regarding npy-matlab the new version has another folder `npy-matlab` so the check is correct.

- for the call this is exactly how it's done: https://github.com/SpikeInterface/spiketoolkit/blob/dev/spiketoolkit/sorters/tools.py#L61-L66

shlex.split splits the command, but not the text betwee `"`, which are the matlab commands. I tested both on linux. The `-wait` argument does not work on linux, so I'll make a check there and add the wait on mac and windows, but not on linux.

alejoe91:
The `-wait` is only needed for windows. So maybe the problem is in the `"`. It's probably a good idea to pass the split path, but I'll make a function in the tools.

PR #44, opened 2019-02-14
Sorter with classes

samuelgarcia:
This PR is the port towards classes approach.

This very first comit is the proof of concept of class approach + one axample with **klusta** only.

@colehurwitz31 @alejoe91 @magland @mhhennig : could you have a look  at least for the API.

I write a SorterBsae with virtual methods and try to implement it on klusta.

**pararlel** and **by_property** is not done at the moment. So it is not too difficult to read.

This is totally untested, just a proof of concept for the API.

Could you give me some feedback ? At least next Wednesday on meeting.

I will next:
  *  make the same for tridesclous and
  * make sorterlist.py that deal with sorter list (unstalled or not installed
  * make several ngine to run sorters in paralelle with thread + multiprocessing in base class.

best

Samuel


alejoe91:
I was able to look at the entire files and I really like the organization :)

I can help in porting the spikeSortByProperty stuff. Let's port the other spike sorters in the next days.

samuelgarcia:
If you like the main idea I can port other sorters very soon.

spiketoolkit/sorters/basesorter.py
```diff
@@ -0,0 +1,116 @@
...
+
+"""
+
+import time
+import copy
+from pathlib import Path
+
+class BaseSorter:
+    
+    sortername = '' # convinience for reporting
```
colehurwitz31: sortername --> sorter_name

colehurwitz31: Just want to be consistent with variable names (this is small)

spiketoolkit/sorters/basesorter.py
```diff
@@ -0,0 +1,116 @@
...
+        assert self.installed, """This sorter {} is not installed.
+        Please install it with:  \n{} """.format(self.sortername, self.installation_mesg)
+        
+        if output_folder is None:
+            output_folder = sorter_name
+        
+        self.output_folder = Path(sorter_name)
+        self.recording = recording
+        self.debug = debug
+        self.by_property = by_property
```
colehurwitz31: by_property --> sorting_property

spiketoolkit/sorters/basesorter.py
```diff
@@ -0,0 +1,116 @@
...
+        self.parallel = parallel
```
colehurwitz31: parallel --> is_parallel?

spiketoolkit/sorters/basesorter.py
```diff
@@ -0,0 +1,116 @@
...
+        
+    def _setup_recording(self):
+        # need subclass
+        raise(NotImplemenetdError)
+
+    def _run(self):
+        # need subclass
+        raise(NotImplemenetdError)
+    
+    def get_result(self):
```
colehurwitz31: get_result should check to see if output file exists or if the spike sorter has been run

spiketoolkit/sorters/basesorter.py
```diff
@@ -0,0 +1,116 @@
...
+        sorting = self.SortingExtractor_Class(self.output_folder)
+        return
+    
+    
+    # new idea
+    def get_params_for_particular_recording(self, rec_name):
```
colehurwitz31: What is this for?

spiketoolkit/sorters/basesorter.py
```diff
@@ -0,0 +1,116 @@
...
+       """
+       this is speculative an nee to be discussed
+       """
+       return {}
+
+
+# generic laucnher via function approach
+def run_sorter_engine(SorterClass, recording, output_folder=None,
```
colehurwitz31: This looks good, but I would call it:

run_sorter_engine --> run_sorter

spiketoolkit/sorters/klusta/klusta.py
```diff
@@ -38,127 +28,113 @@ def klusta(
...
+    def __init__(self, recording=None, output_folder=None,
+                                    by_property=None, parallel=False):
+        BaseSorter.__init__(self, recording=recording, output_folder=output_folder,
+                                    by_property=by_property, parallel=parallel)
+
+    def set_params(self, **params):
+        self.params = params
+        
+    
+    def _setup_recording(self):
```
colehurwitz31: I don't understand what _setup_recording does to be honest, 

- Why does it save a probe file and also save out to a binary format?

spiketoolkit/sorters/sorterlist.py
```diff
@@ -0,0 +1,11 @@
+
```
colehurwitz31: Is this file just keeping track of all sorters that are installed?

spiketoolkit/sorters/tridesclous/tridesclous.py
```diff
@@ -1,95 +1,125 @@
...
+        initial_catalogue = self.tdc_dataio.load_catalogue(chan_grp=0)
+        peeler = tdc.Peeler(self.tdc_dataio)
+        peeler.change_params(catalogue=initial_catalogue,
+                            use_sparse_template=use_sparse_template,
+                            sparse_threshold_mad=1.5,
+                            use_opencl_with_sparse=use_opencl_with_sparse,)
+        peeler.run(duration=None, progressbar=True)
+
+
+def run_tridesclous(
```
colehurwitz31: So this is a different function then run_sorter_engine?

colehurwitz31: Should they be the same?

colehurwitz31:
Looks great! I think it should be how we move forward with spike sorting algorithms

I did a little code review and had some questions about naming and functions. It seems like the run_sorter function for trideclous and the basesorter are different?

spiketoolkit/sorters/tridesclous/tridesclous.py
samuelgarcia: At this moment it is the same.
I was wondering between a unique generic one or one runner function for each SorerClass.

The run_sorter_engine have also in the signautre the SorterClass itself.

the run_tridesclous have the same signeture as beefore (and so for run_klusta)

spiketoolkit/sorters/sorterlist.py
samuelgarcia: It is a list for all sorter and all installed sorter.

So for testing sorters on a particular dataset the list will generate automatiocally with future helper fuction.

spiketoolkit/sorters/klusta/klusta.py
samuelgarcia: It is split in 2 steps:
  * setup recording (basically copy/paste files), it can take long time for big files
  * and the comptation itself.

If we want to make benchmark on the second step only it will be usefull.

spiketoolkit/sorters/basesorter.py
samuelgarcia: OK

spiketoolkit/sorters/basesorter.py
samuelgarcia: In my mind, some sorter will need to adapt the default parameters depending on the dataset caracteristics. Sorters with same parameters don't behave the same depending if the dataset is a tetrode or dense array. in vitro or in vivo, etc...
So for a fair comparison, it should be possible to change parameters for a given dataset label.
It is just a futur proposal.

spiketoolkit/sorters/basesorter.py
samuelgarcia: Yep.
Good idea.

spiketoolkit/sorters/basesorter.py
samuelgarcia: yes of course. a typo.

spiketoolkit/sorters/basesorter.py
colehurwitz31: I see. Maybe a function like, "adapt_parameters(self, recording)" which takes in a recording extractor and allows the algorithm to adapt its default parameters. We don't want to pass in the recording_file_name because we are trying to abstract away the file type for that.

spiketoolkit/sorters/klusta/klusta.py
colehurwitz31: So set_up recording does what exactly? I thought that the recording extractor was supposed to eliminate the need for copying over big files. Am I missing something?

spiketoolkit/sorters/sorterlist.py
colehurwitz31: I see. Sounds good.

spiketoolkit/sorters/tridesclous/tridesclous.py
colehurwitz31: I think each algorithm should have the same function, "run_sorter". The name and signature for this function should be the same across spike sorters. Is that possible? It would make running all sorters the same which is great for standardized analysis.

spiketoolkit/sorters/basesorter.py
samuelgarcia: OK

samuelgarcia:
Yes idealy run_sorter can run all SorterClass but by compatibility I also did run_tridesclous and run_klusta that have same signature as tridesclou() and klusta().

So all examples only need to change klusta(..) to run_klusta(...)

colehurwitz31:
I would rather change all the examples to run_sorter. We don't have many examples to fix.

samuelgarcia:
I also prefer.

samuelgarcia:
I am trying to move the moutainsort4 function to class.

There is something different for this one : there is no extractor in spikeextractor.

The ml_ms4alg directly render a sorting with its own input ouput function.
This could limit a bit the new class approach that decompose (for multiprocessing reason) in 3 steps: 
  * setup recording
  * run
  * get spikeextractor


What is the reason of having no extractor in spikeextractor of this sorter ?

samuelgarcia:
Other problem in moutainsort4, there is no output_folder.

This will make the framework difficult because there is no explicit persistency on disk.

In my mind, for multiprocessing, this should be mandatory to all sorter.

colehurwitz31:
I am not sure what you mean, in the mountainsort algorithm, a recording extractor is used throughout (each recording extractor is called a recording). Also, it uses sorting extractors (each sorting extractor is called a sorting).

We can look into adding an output_folder to the sorter. Maybe Jeremy can weigh in on that. @magland

spiketoolkit/sorters/klusta/klusta.py
alejoe91: hei @colehurwitz31 
many of the spikesorters require to save the data in a specific format (e.g. dat) and or provide probe files and params files. 

The _setup_recording() is a nice way to organize these steps.

Thanks @samuelgarcia

alejoe91:
@samuelgarcia and @colehurwitz31 
I can comment on the mountainsort approach, as I'm also implement the same idea in SpyICA (https://github.com/alejoe91/spyica).

As we have the SpikeInterface framework implemented, the idea is to bypass the spike-sorting specific formatting etc, so that the spike sorting functions themselves directly work on the recordingExtractor object and return a sortingExtractor one. 

@magland It would be nice to have ms4 alg to accept an output_folder argument instead of using the /tmp/ folder by default. I have made some changes to ml_ms4alg and can make a PR. 

@samuelgarcia for now I would put an dummy `output_folder` arg for standardizing the inputs.
A possible solution could be having a second Base class, `BaseSISorter` (where SI is SpikeInterface) for sorters which are developed within the SpikeInterface framework.

What do you guys think?

colehurwitz31:
I think the plan should be for all sorters to take in a recording extractor and output a sorting extractor (can also save a normal output folder). 

I think the BaseSorter class should be sufficient for this purpose rather than defining two classes.

alejoe91:
@colehurwitz31 I think you are missing the point. We are talking about adhering to the _setup_recording(), run(), get_sorting() implementation.

For mountainsort, this is not possible because the ms4 function from ml_ms4alg package already takes a recordingExtractor and returns a sortingExtractor. So the _setup_recording and get_sorting functions are not needed

colehurwitz31:
Wouldn't it be:

1) _setup_recording() --> Do nothing (or whatever setup is needed to use ml_ms4alg)
2) run()  --> stores the sorting extractor internally after running ml_ms4alg
3) get_sorting() --> returns the internally stored sorting extractor

I don't see the issue with this. Every algorithm will do something similar and if they don't use one of the steps, that is ok.

samuelgarcia:
Another keypoint iwith this new approach is the ability to serialize a sorter (so its parameters and output_forlder) to be executed elsewhere (subprocess or other machine).
So ideally, even if it is not the case now. The python interpreter could stop and restart in bewteen theses steps to facilitate batch across process and then collect results.

In short, a sorter should not have to resident in memory for further analyses.

samuelgarcia:
@alejoe91 : I would be more in favor of having one unique class.
Another class BaseSISorter would make the framework more fuzzy.


My feeling is that sorting tools that already have a "SpikeInterface" API builtin,  should **also** have a spikeextractor class even if it is a trivial  function to function mapping.

So the get_result would be more generic.

We should really have in mind that comparing several sorter on real big fat datasets (>256 channel over minuts and more) will never fit in memory even the results.

alejoe91:
Maybe we could extend the NumpyExtractor so that it could load map numpy files?

Than we could just dump the unit_ids, spiketrains, and possibly properties and features and load them from file?
It would be nice to have the same for recording extractors too.

alejoe91:
BTW, I can help implementing some sorting classes. Should I tackle kilosort?

colehurwitz31:
> Maybe we could extend the NumpyExtractor so that it could load map numpy files?
> 
> Than we could just dump the unit_ids, spiketrains, and possibly properties and features and load them from file?
> It would be nice to have the same for recording extractors too.

spiketoolkit/sorters/ironclust/ironclust.py
```diff
@@ -1,7 +1,163 @@
...
+        'detect_threshold': 5,  # Threshold for detection
+        'merge_thresh': .98,  # Cluster merging threhold 0..1
+        'freq_min': 300,  # Lower frequency limit for band-pass filter
+        'freq_max': 6000,  # Upper frequency limit for band-pass filter
+        'pc_per_chan': 3,  # Number of pc per channel
+        'parallel': True
+    }
+
+    installation_mesg = """
+       >>> pip install mountainlab_pytools
```
alejoe91: can you add: `kbucket` to the pip install? That is also needed for ironclust

samuelgarcia:
@alejoe91 : I have push some minut ago teh kilosort class.
But I will need a lot of help for testing theses code.

I don't have matlab installed on my mahcine so kilosrt and ironclust is not tested at all!!!!!
I did not run then.
What you can do is make a branch from my repo with:
 **git remote add samuelgarcia https://github.com/samuelgarcia/spiketoolkit.git**
Then, debug ironclust an kilosrt.
And then push back via a PR on my repo.

I am doing the circus.

spiketoolkit/sorters/ironclust/ironclust.py
samuelgarcia: with pip install kbucket ?

alejoe91:
Just saw it :)

Ok I will update the run_all_sorters notebook and make sure that everything runs. Will also add tridesclous

samuelgarcia:
At the time right now only tridesclous is working ...
klusta should also work but I have "Exception: Klusta did not run successfully"
due to bad execution.

You can have a look in spiketoolkit/sorters/test and try to run then one by one.

alejoe91:
@samuelgarcia 
How is the installation_mesg supposed to work? can we make it self in the constructor?

klusta, ironclust, and kilosort run. Just wanted to figur out this before making a PR

alejoe91:
Regarding my previous comment: I see it's static already! Haven't used static variables before in Python and had to google it...

alejoe91:
@samuelgarcia 
made a PR for klusta, kilosort, and ironclust fixes

colehurwitz31:
Hi all, 

Is this ready to be merged? If not, is there anything I can do to help?

Best,

Cole

samuelgarcia:
No.
This is not ready.

We need to:
  * make the parralel and by properties thing.
  * debug classes
  * progagate changes everywhere (doc, examples)
  * make the moutain sort class working.
  * make travis working

I will try at the end of the weekd and will need help from Alessio from the by property.

At the moment, to help you can clone this and test some classe randomly.

Look at sorters/tests/test_*.py

colehurwitz31:
So just to clarify, I would clone and then try running each of the tests by installing spike sorters? Then, I should report the results to you somehow?

samuelgarcia:
You can even directly debug it on your report with a git remote add and make a PR on my report before the "big" merge into main report.

samuelgarcia:
I dig a bit in moutainsort4 code.

inside ml_ms4alg module moutainsort4 function is this one:

```python
  MS4=MountainSort4()
  MS4.setRecording(recording)
  geom=_get_geom_from_recording(recording)
  MS4.setGeom(geom)
  MS4.setSortingOpts(
    clip_size=clip_size,
    adjacency_radius=adjacency_radius,
    detect_sign=detect_sign,
    detect_interval=detect_interval,
    detect_threshold=detect_threshold
  )
  tmpdir = tempfile.mkdtemp()
  MS4.setNumWorkers(num_workers)
  print('Using tmpdir: '+tmpdir)
  MS4.setTemporaryDirectory(tmpdir)
  try:
    MS4.sort()
  except:
    print('Cleaning tmpdir:: '+tmpdir)
    shutil.rmtree(tmpdir)
    raise
  print('Cleaning tmpdir::::: '+tmpdir)
  shutil.rmtree(tmpdir)
  times,labels,channels=MS4.eventTimesLabelsChannels()
  output=se.NumpySortingExtractor()
  output.setTimesLabels(times=times,labels=labels)
```

I propose to base the class Mountainsort4Sorter on that piece of code instead of having this output_input fixed function.
So it could fit more the new class approach that decompose in several steps and fix the output_folder thing. @magland @alejoe91 : is it OK ?

I do proposal soon.

samuelgarcia:
There is still one problem with moutain sort is that there is no extractor directly in spikeextractor.
I think we should write one based on mountainlab_pytools.
@magland : what do you think ?

samuelgarcia:
I am about to make the _spikeSortByProperty.

I think I understand the thing.
To be clear. If property is not None then a list of SubRecordingExtractor is done.
And so a list of sorting is computing in parrlell via thread or not.

I plan to implement this as follow in BaseSorter.
recording will always be a list recordings event there is only one.
Internaly get_result**s(..)** will a be a list of SortingExtractor and get_result() will make a MultiSortingExtractor in any case.

My my concern with that approach is that there are several way to make several "groups".
This actual one with "property" and the second ihhen one with the probe file.
I think that the 2 approach should converge, in short "probe_file" should be a common keyargs even if the sorter do not deal with it and this probe file should be used to create a "group_id" property for channels taht would be reflected in the recodring so the SubRecordingExtractor would also work inthat case.

What do you think ?

alejoe91:
Everything looks good but I thin that the 2 apporaches to spikesort by group are already convergent. 
When you load the probe file (or pass it as an arg - which we could make a common keyarg) the group information (if present in the probe file) is loaded in the recordingExctractor as channel property. 
Then one can call the spike sorting function with propert=group. This should automatically:
- split the recording extractor in several subrecordings based on the property (done with se.getSubExtractorsByProperty)
- run the spike sorting in parallel
- reassemble the results and adding the right property to the unit (e.g. unit 23 is in group 6)

this approach allows to deal with the probe_files (in that case the property would be group in most cases), but it also keeps it flexible in case the user wants to split the sorting by other means.

Is it clear?

samuelgarcia:
OK. This was the piece I totally miss.
Good I anderstand everything. I will implement this soon so.

samuelgarcia:
Is it OK if I change the "by_property" to "grouping_property" ?

samuelgarcia:
Is it OK to add this behavior:

* when there is an explicit probe_file with several group AND grouping_property is None then the
  grouping_property is automatically set to "group" ?

samuelgarcia:
There is a conflict.
It is more complicated than I though.

For sorter based on PRB file (tridesclous, circus, klusta, ...) the input trace (recording) have all the channel. Inside the prb file each group is mapped with the absolut channel map.
If we use the "group" property for theses sorters, then traces are reduced to a smaller subset of channel count and then the PRB is wrong because it play with absolut channel index.

In short, sorter with no PRB file support the grouping is done **externally** by the SubRecordingExtractor but for sorter than support PRB file the grouping is done internaly by the sorter.

So we need that if probe_file is not None then "by_property" (or "grouping_property") must be None.

Am I clear ?

alejoe91:
Hi Sam,

I don't think I understood the issue here, but the way I use it now the probe is loaded externally and the spike sorter is run. 
I think this is a good way to do it: load the probe file before running sorting:
- load extractor
- load probe file
- run spike sorting

samuelgarcia:
@alejoe91 
I code the proof of concept of multi group handling in BaseSorter.

I change a bit the API.
Now the setup, run and result is done explicitly on one rescording (or subrecording).
So:
  *  **def _setup_recording(self, recording, output_folder):**
  *  **def _run(self, recording, output_folder):**
  * **def _get_one_result(self, recording, output_folder):**

With this it is easy to loop over subrecording for multi groups and launch them in parallel.

I push changes only for tridesclous.

If you are happy with this I will make change for all sorters.

samuelgarcia:
Also note that I start to make a SorterCommonTestSuite class that will launch a serie of test for each sorter class with a very simple hineritence system.

So basic feature of sorter will tested for each one with pytest when installed.

samuelgarcia:
3 task to do before merge:
  * make the moutain sort class working (I need help here)
  * make travis working
  * debug. Only tridesclous pass test on my machine.

samuelgarcia:
Yeah!!!!
Travis is passing test!!!!
🥂

Situation of CI for sorter:
  * klusta : tested and pass on travis
  * tridesclous :  tested and pass on travis
  * circus: hard to install on travis because the MPI
  * kilosort : never will run travis because matlab
  * ironsort : never will run travis because matlab
  * ms4 : no implemented yet but should ru easly on travis.

@colehurwitz31 @alejoe91 : could make a full pytest on you mahcine if you have all sorters installed ?

@magland @alejoe91 : still need help for MS4

colehurwitz31:
Wow!! Nice work :) I think @alejoe91 managed to install all the spike sorters (with some dark magic) so many he can quickly make a full pytest.

samuelgarcia:
> with some dark magic
+ a matlab licence :(

alejoe91:
Wow @samuelgarcia this is amazing!

I made a PR with a working implementation ( a bit hacky) of MS4 in this framework.

One think that we should add is the ability to provide only a few arguments. If a parameter is not given then the default is used. Right now ALL the parameters need to be passed. Maybe a check_parameters in the base class that loop through the keys of _default_params and kwargs and uses the default one if not ofund in kwargs?

alejoe91:
> * circus: hard to install on travis because the MPI

I think mpi can be easily installed as follows from the .travis before the install:

```
sudo: enabled
addons:
  apt:
    packages:
    - libopenmpi-dev
    - openmpi-bin
```

I can try to test if spyking-circus will run with this plus pip install spyking-circus

samuelgarcia:
OK I will try circus on travis.

The fact is that circus is buggy on my clean install because in some situation circus try to open a GUI that close immediatly. And it create problem... I guess on github it will be worse.

Have you test circus locally ?

samuelgarcia:
For the parameters, Iwanted but I forgot to make a set_params at BaseSorter level that take all default_parms and update with given one. It was my plan but I forgot it.

Ithink it will fix what you say for params.

samuelgarcia:
Parameters fixed. Thanks.

magland:
Hi all. I just wanted to give brief feedback. It seems like you are doing
very good work. I guess my intuition tells me to keep it simple with
function calls and parameters. But I see the advantage of putting all
sorters in the same framework with base class. One could always create a
simple interface on top if needed.

samuelgarcia:
Hi @magland .
Thank you.

The previous functional interface is  kept with but **run\_kilo(..)** instead of **kilo(..)**, **run\_klusta(...)**. instead of **klusta(..)**, ...

Have a look in sorterlist.py, it is quite simple.

magland:
Nice!

alejoe91:
@samuelgarcia I'm running all tests and will push a commit on the PR on your fork as soon as they all run (also on travis)

samuelgarcia:
Travis pass test but it fact it run only a few of then. (only tdc and klusta)
circus and ms4 shoudl also run there.
Could someone could clear the cache of travis ?

samuelgarcia:
I need some help.

I am fighting with pytest on travis.

The travis version of pytest is 3.3.0. this version is too old and do not deal well with class unittest inheritence.
So, no test are launch except a few one.
The script do pip install pytest and so the version should 4.3.0
With this version tests behave better.

I tried a python -m pytest but it do not change anything.


Any ideas ?

alejoe91:
@samuelgarcia 
Is the problem on travis or locally?

alejoe91:
you can try pip install pytest==4.3

samuelgarcia:
locally I have virtualenv so I am sur that the pytest is the good one.

On travis, my guess is taht pytest is already installed at system level but it is an old one.
So th "pip install pytest" which install the latest version have no effect.

On my machine, I am able to run tests and crash then :)
On travis everything is skipped because of the design I choose (class inheritance  with SorterCommonTestSuite)

alejoe91:
I see. Have you tried pip install pytest==4.3 from the .travis?

alejoe91:
@samuelgarcia the travis stuff is fixed in the new PR i made to your fork :)

samuelgarcia:
I will be off at the begining of the next week.

If you are happy with I think we can merge this PR. [if travis pass all test]

I will be back on launcher.py soon with a new PR.
The actual on is just a proof of concept.

alejoe91:
What do you guys think to have a dev branch with all these major API changes and merge when everything is done and stable? @samuelgarcia @colehurwitz31 @magland

samuelgarcia:
My opinion in a fresh and new project like this one is to merge everything and soon to master to avoid several branches.
You can put tags sometimes instead if you want retro compatibility.

Figuring which feature is in the dev and wich is in the master. is really time consuming.

You should consider the master (as the dev) + branch in private repo for feature + tags(to release). No more.

alejoe91:
Then let's wait to merge as we use extensively spiketoolkit in our analysis pipeline. Have to update and test the other codes too :) Give me a couple of days

colehurwitz31:
How exactly the class system can help with dependencies between spike sorters? I remember we discussed this, but I am still unclear.

Do you think it would be a good idea to make docker images for each spike sorter? That could be a way to handle all the dependency issues.

alejoe91:
@samuelgarcia 

made a PR to you fork with some changes. After that we can merge here :)

samuelgarcia:
@alejoe91 : I made some comment in the sub PR.

samuelgarcia:
@colehurwitz31 : at the moment there are no dependency "between" sorter.

For each sorter class there is a ```installed``` attribute that can be check to see if the sorter if usablme or not.
See sorterlist.py.
The ```installed``` mainly depend for each sorter of a global variable that is initialized at module level with help of ```try .. except ModuleNotFoundError```

For instance in tridesclous:
```
try:
    import tridesclous as tdc
    HAVE_TDC = True
except ModuleNotFoundError:
    HAVE_TDC = False

class TridesclousSorter(BaseSorter):
   ...
    installed = HAVE_TDC
    ...
```

In basesorter if installed=False then an error is raised an the ```installation_mesg``` is show to the user.


For a total automatic installation of each sorter we could also add a classattribute like
 ```
class TridesclousSorter(BaseSorter):
   ...
    install_cmd = """
pip install https://github.com/tridesclous/tridesclous/archive/master.zip
""""
 ```

At the moment the command is in the ```installation_mesg```

PR #59, opened 2019-03-01
[WIP] Several new features

samuelgarcia:
Do not merge for now.

Several features:
  * Save raw file in tridesclous less memory agressive.
  * laucher save run time for each sroter/recording in csv

### DIFF
--- a/spiketoolkit/sorters/basesorter.py
+++ b/spiketoolkit/sorters/basesorter.py
@@ -22,6 +22,7 @@ import copy
 from pathlib import Path
 import threading
 import shutil
+import os
 
 import spikeextractors as se
 
@@ -56,12 +57,15 @@ class BaseSorter:
             # several groups
             self.recording_list = se.getSubExtractorsByProperty(recording, grouping_property)
             n_group = len(self.recording_list)
-            self.output_folders = [Path(str(output_folder) + '_'+str(i)) for i in range(n_group)]
+            if n_group>1:
+                self.output_folders = [Path(str(output_folder) + '_'+str(i)) for i in range(n_group)]
+            else:
+                self.output_folders = [output_folder]
         
         # make folders
         for output_folder in self.output_folders:
             if not output_folder.is_dir():
-                output_folder.mkdir()
+                os.makedirs(str(output_folder))
         self.delete_folders = delete_output_folder
 
     @classmethod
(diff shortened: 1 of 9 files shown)


samuelgarcia:
@alejoe91 : I would like to change a bit the BaseSorter.

As you known, I would like to have a static method to read result from forlder.
So we could open result from intermediate folder after a long run.

Ideally, it would be a staticmethod like this
```python
sorting_KL = KilosortSorter.read_from_folder('my_output_folder_KL')
sorting_TDC = TridesclousSorter.read_from_folder('my_output_folder_TDC')
```
I need to patch the MS4 as you known.

I also need a change in kilosort, circus and klusta.
These sorter have "file_name" that is the name of the internal name inside the folder.
By default it is recording for all of then.
My guess is that this file_name is not set because it do not matter at all because each sorting is always done in separte directories.

This "file_name" for this sorter (kilosort, circus and klusta) is also a instance attribute done in **_setup_recording** and so it is needed for **_get_one_result**

To implement the staticmethod **read_from_folder** I need to get rid of this file_name.

Is it possible to remove it and to generate always the deault one ("recording") ?

alejoe91:
@samuelgarcia 
Yes I think we can get rid of file_name and use `recording` by default. Then output folder will be the only thing we need.

Regarding MS4 I think an easy fix would be to change:
`self._sorting_result[ind] = sorting`

to

`se.MdaSortingExtractor.writeSorting(sorting, str(output_folder / 'firings.mda'))`

then the _get_one_result will become:
```
    def _get_one_result(self, recording, output_folder):
        sorting = se.MdaSortingExtractor.(str(output_folder / 'firings.mda'))
        return sorting
```

This should make it compatible with the class organization and allow the read from folder. Right?

samuelgarcia:
Cool. I will do.
For MS4, it was the plan already.

samuelgarcia:
@alejoe91:
My week will be very busy. I won't have time to work more on this.
You should merge this because it change a bit the class BaseSorter.

I will be back on launcher next week.

I think you will have to make some changes in KS2.

alejoe91:
I'll test it first and the n merge. I have to check if it breaks our code and make the changes in case

Alessio

samuelgarcia:
Of course.

PR #61, opened 2019-03-05
kilosort2 init

alejoe91:
- it runs on my machine but at one point I get CUDA_ERROR_ILLEGAL_ADDRESS

- probably needs an updated version


alejoe91:
add tests

PR #64, opened 2019-03-11
Add herdingspikes support

martinosorb:
This seems to work, at least for me, and we can discuss together.

What I would like to do is to remove some of the default parameters so that the user is forced to set them manually, but I'm not sure what's the best way to do this.

### DIFF
--- a/spiketoolkit/sorters/herdingspikes/__init__.py
+++ b/spiketoolkit/sorters/herdingspikes/__init__.py
@@ -1,0 +1,1 @@
+from .herdingspikes import HerdingspikesSorter
--- a/spiketoolkit/sorters/herdingspikes/herdingspikes.py
+++ b/spiketoolkit/sorters/herdingspikes/herdingspikes.py
@@ -1,0 +1,135 @@
+import os
+import shutil
+
+from spiketoolkit.sorters.basesorter import BaseSorter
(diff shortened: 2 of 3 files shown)


alejoe91:
Hi @martinosorb 

Thanks for the addition. It looks good but I recommend changing a couple of things:

- `_default_params` : it's better to have it as a flat dictionary (not nested). Just expose the most useful parameters for the moment
- from the latest  PR, sorters need to implement the staticmethod `get_result_from_folder` (see example here: https://github.com/SpikeInterface/spiketoolkit/blob/master/spiketoolkit/sorters/klusta/klusta.py#L114-L117)

Could you add those?

Thanks

martinosorb:
I've added the method, which should work if the sorting extractor works. I copied the nested dictionary structure from tridesclous, it's useful because I can avoid calling all parameters one by one as kwargs. What do you mean by "just expose the most useful ones"?

alejoe91:
Regarding the parameters I think it's usefult to have a flat dictionary with the most useful/user-tunable params and then build the spike sorting dependent params structure in the function. But I guess this can be done later :)

Merging now

martinosorb:
Thanks for merging, I see what you mean. We'll have to decide which parameters need to be exposed with Matthias

Issue #71, opened 2019-03-18
Enhancement in sorter/launcher and avoid copy in sorter


samuelgarcia:
Many sorter have by internally the RawBinary as input.
In spiekextractor, it is the BinDatRecordingExtractor with frames_first=True.

This is more or less the universal format.
The main drawback, is that it do not have any metadata (dtype, nb_channel, smaple_rate,...)

Many sorter use this : tridesclous, spyking-circus, kilosort, kilosrt2, klusta, yass.

In the actual way all sorter are implemented, to make possible to have any extractor as input, each sorter make a deepcopy inside the **output_folder** with **writeBinaryDatFormat**.

For big files (like the Sampinato ground truth), this make the run_time totally unefficient, because the input file is copied many times.
Further more the writeBinaryDatFormat load everything in memmory to achieve this, (except for tridesclous where I made a by chunk writeBinaryDatFormat).

I propose the following change for theses sorter :
  * if the recording is **BinDatRecordingExtractor** and **frames_first=True**, then the file would 
    not be copied and the **BinDatRecordingExtractor._datfile** would be give to the sorter to 
    avoid this deepcopy.

So in sorter/launcher.py, I would also propose this change:
   * for each recording, force a **writeBinaryDatFormat** in the working_folder for each recording.
     Create a new recording with BinDatRecordingExtractor and give this input to all sorter.

With this we could solve many copy on the drive and we could facilitate to run many sorter on a cluster because only the rawbinary file path would be given which is easy to serialize (as oposed to Recording).

samuelgarcia:
Started in #82

alejoe91:
this has been merged. Closing!

Issue #69, opened 2019-03-15
Proposal about spiketoolkit.comparison


samuelgarcia:
Today I have discovered and explore a bit the comparison submodule.
I mainly play with the sorting comparison at the moement.
This is very cool and cover many needs.
Thank you.

Here some feedback about the code itself:
  * There is a lake of documentation
  * there is a lake of unitests
  * variable do not have obvious naming for a newcomer (TP, FP, gtst, sst).  - BTW what is CL ? -

The class approach for SortingComparison is very good from the user behavior view.

But internally important and complicate function like **_do_matching _do_counting _do_confusion** are hard to read because they are in the class itself.

Theses 3 functions, as they are complicated should come with many unitests.
I think having this functionality inside isolated function would be better for the code debugging and reading. We could do that without breaking the behavior the method of class would do a call to the function.

@alejoe91 @colehurwitz31 @mhhennig : would it be OK if I would propose some refactoring on this module (100% compatibale of course) with this implementation that would split the difficulty of testing each piece ? I would also include some docs and variable renaming.

I have another proposal about this submodule:
  * **use pandas as backend to render results**

Some function like **compute_performance** are cool to display one results.
But we could the same function for a **list** of comparison and we would have a DataFrame with index as comparison and mletrics as columns.
What do you think ? It add a depency for the project but it is a very common one in the python world.

It would be perfect to compbing the laucher sorter on many datasetXsorter and the fact that we would collect many metrics in one function that return a  dataframe.
What do you think ?

I would be happy to hear your feedback about this.

alejoe91:
Hi @samuelgarcia 

I think splitting the function that need tests in separate functions, maybe in a sorting_tools file is a great idea. I agree we should write extensive tests on them, e.g. comparison between the same set of spike trains etc.

Documentation should also be added.
BTW, CL is classification error - gtsts: ground truth spike train - sst: sorted spike trains. This is taken from literature, but I agree that it's not straightforward and it should at leat be documented.

I'm also ok to use pandas as backend. I think that spikewidgets already has pandas as deoendency, and as you sais, it's a very common one.

Do you want to give it a try at refractoring? I can help too :)

alejoe91:
Also I think that plotConfusionMatrix should be ported to spikewidgets

colehurwitz31:
Yes, I agree with your comments Samuel! Thanks for looking at each of our modules with a critical eye, it is very helpful.

[...]

Issue #65, opened 2019-03-12
Add test_heardingspikes.py in unitest (st/sorters/tests)


samuelgarcia:
Add test_heardingspikes.py in unitest (st/sorters/tests)

samuelgarcia:
Done in #73 but have some travis problem now.

PR #73, opened 2019-03-18
Refactoring sorting comparison module + pandas

samuelgarcia:
Several features:
  * test for HS2
  * gather SortingSomparison taht works with launcher in one function to create dataframe easily
  * split complexity of SortingComparison into separated function in comparisontools.py
  * make example of the usease with pandas


samuelgarcia:
This is about:
  * #65 
  * #71 
  * #69

samuelgarcia:
I push an example [here](https://github.com/samuelgarcia/spiketoolkit/blob/new_features/examples/sorter_comparison.ipynb)
of the behavior I would like to have with the new sorters.launcher and comparison.gather_sorting_comparison with pandas.

The internal implementation is not done yet I need to modify SortingComparison to make it more firendly with pandas but it give an idea.

@alejoe91 @colehurwitz31 @mhhennig : could you give me some feedback of the external API with pandas ? Do you have more ideas ?

colehurwitz31:
I think the External API looks great! I am not to familiar with comparing sorters, but the metrics and tables you have seem like a good start. I can try to think of more ideas.

samuelgarcia:
OK.
I will start today the refactoring.

samuelgarcia:
I am reading the sortingcomparison code to port it into function to make it easier to debug.

the task will harder than I thought, there are many internal call to small function like theses

  * getBestUnitMatch1, getBestUnitMatch2

    That can be avoid with some simple python dict function like
    ```
       def getBestUnitMatch1(self, unit1):
           if unit1 in self._best_match_units_12:
               return self._best_match_units_12[unit1]
           else:
               return None
    ```
    is 
    ```
    self._best_match_units_12.get(unit1, None)
    ```

  * getMappedSorting1 getMappedSorting2 use a class MappedSortingExtractor.


All theses sub methods, make the code hard to follow/anderstand.
I thaink that I will need more help @alejoe91 are you volunteer ?


Another question : 
what is difference between _best_match_units_21 and _unit_map12 ?
From what I anderstand _best_match_units_21 is max between counting and _unit_map12 come form the Hugarian algo.

alejoe91:
That was mainlywritten by @magland, but I think I can help out :)

- use `self._best_match_units_12.get(unit1, None)` sounds great (I didn't know)

- the MappedSortingExtractor is derived from a basic SortingExtractor. The idea is that you can use functions/widgets on the mapped extractor (example plotting waveforms and mapped waveforms side by side)

- we can also ask @magland and have a chat about the implementaiton

- your guess is right. best match is the max count and unit_map12 is the output from the hugarian algo (1 to 1 match)

samuelgarcia:
I have just put a new commit with separate comparisontools.py.

I made 3 functions:
  * get_matching
  * get_counting
  * get_confusion

I made a basic signature + doc.
I would need help to make clear docstring it would help me for making this function out the main class.

In my mind theses 3 function would not depend to much on subfunctino/metyhods to make the code easier to read for a new comer.

samuelgarcia:
@magland : would you OK to help to port theses 3 functions outsie the main class to improve the code readability ?

samuelgarcia:
I have started a try to port do_matching to pure function see comparisontools/get_matching.
Ready for feedback.

samuelgarcia:
I continue to port SOrtingCOmparison into splited function.

Could you confirm that:

```self.getMappedSorting1().getMappedUnitIds(u1)```
is equivalent to
```self._unit_map12[u1]```
when u1 is scalar.

**Comment by alejoe91** [comment]
Yes it shuould be the same :)

**Comment by samuelgarcia** [comment]
Thank you.

@magland @alejoe91 
So I have ported matching, counting and fonfusion into indepeante function.
The reading is much simpler because there are no call to other function.

I have also made simple tests.

Could you check this ? comparison/comparisontools.py.

if this is OK, I will be able to make the big refactoring inside SortingComparison.

thanks.

**Comment by alejoe91** [comment]
@samuelgarcia the comparisontools.py looks good to me.

Can you do the refractoring here or do you prefer to merge this first?

**Comment by samuelgarcia** [comment]
Please do not merge. If you validate the comparisontools.py, I will propagate changes into SpikeComparison too.
And fix theses travis errors.

Even if comparisontools.py look good, I think a carfull read of the whole code comparing it to old implementation should be good. As changes affect man lines, I would sorry to introduce errors.

And also, I will need help (in sub PR) do make good docstring for each function, with 
a strong emphasis on variable name exaplanation. My english is so bad.

In my mind, this SortingComparison is really crucial to make this tools adopted by the community.
I guess that other that me will want to inscpect theses performance metrics so the doc need to be cristal clear. No ?

**Comment by samuelgarcia** [comment]
One more question in actual SortingComparison there is a function
**def confusion_matrix()** with a strange signature.
My guess is that it is an old implementaion because teh actual is in **_do_confusion()**.

Must I remove it ? Or is it here for something ?

**Comment by samuelgarcia** [comment]
I have finish the internal refactoring of SortingComparison, the main difficulty is now outside (in comparisontools).

The external behavior of the class is now totally unchanged (unless a mistake).

After this will validated, I would propose some slight change in the class itself:
  * compute_performance as method of the class (but keep func too)
  * move some private attribute and method (starting with **\_**)  to not none private like _sorting1
    to sorting1. I have the feeling that the use of "_" is a bit too much here. Normally this underscore is 
   usefull to tell user "do not use externally" but many of attributes are indeed usefull (self.\_counts, ...)
   externally.
  * remove plotting
  * many more.

**Comment by samuelgarcia** [comment]
I made an enormous mistake!!!
I did not include @mhhennig changes before the refactoring.
I wassure to have done it.

Oups. I stringly need to port then all manually now...

@mhhennig : will you be able to help me byt reading **comparisontools.py** in this PR.

Really sorry.

**Comment by samuelgarcia** [comment]
OK I correct this mistake and copy back Matthias changes in comparisontools.py.

This is ready for review and ready to recieve help on docsting.

**Comment by samuelgarcia** [comment]
When reviewing code : could you reply to "# Samuel EDIT" inside the code ?

**Comment by samuelgarcia** [comment]
I am unlucky with travis.
I have also push here a test_heardingspike.py that is now not working.
I think we don't care to much for the moment, it is a story about [reading hdf5](https://travis-ci.org/SpikeInterface/spiketoolkit/builds/509116774#L1176)

**Comment by colehurwitz31** [comment]
Working on this now!

**Comment by colehurwitz31** [comment]
That was painful. It turns out that the HS2 sorting extractor is memory mapped and leaves the file open. If you don't delete the sorting extractor in the for loop of the test, then you will run into a h5 complaint that you are trying to overwrite a file that is open. I modified the test because I think that it is ok that the sorting extractor leaves the file open.

**Comment by samuelgarcia** [comment]
@colehurwitz31 : thank you very much for the fix.
I don't anderstand why the sorter was trying to overwrite an h5 file in get_result but the fix is OK for me.

Before merge, @colehurwitz31 or @magland should carrefully validate the refactoring because they are original author or spikesomparison.

**Comment by alejoe91** [comment]
I will try to test it today :)

**Comment by samuelgarcia** [comment]
No urgency.

**Comment by colehurwitz31** [comment]
@samuelgarcia since we run each sorter with parallel to false and parallel set to true, we end up overwriting the output file in that for loop. That is completely fine, however, when we made a sorting extractor for hs2, we open up the output h5 file and leave open. Therefore, when hs2 tries to sort again in that for loop, it can’t overwrite the same file because the sorting extractor has it open.

**Comment by colehurwitz31** [comment]
@samuelgarcia: I never did much work on the sorting comparison so I am not sure how much I can add documentation and check if everything is alright. I think @alejoe91 and @magland did most of the work on this.

File: spiketoolkit/comparison/comparisontools.py [hunk 1]
```diff
@@ -0,0 +1,398 @@
...
+            lab_st2 = labels_st2[u2]
+            mapped_st = sorting2.getUnitSpikeTrain(u2)
+            # from gtst: TP, TPO, TPSO, FN, FNO, FNSO
+            for sp_i, n_sp in enumerate(sts1[u_i]):
+                matches = (np.abs(mapped_st.astype(int)-n_sp)<=delta_tp//2)
+                if np.sum(matches) > 0:
+                        lab_st1[sp_i] = 'TP'
+                        lab_st2[np.where(matches)[0][0]] = 'TP'
+        else:
+            # Sam garcia note : I guess this is a mistake
```
alejoe91: This tags all unmatched spike trains as False Negatives

spiketoolkit/comparison/comparisontools.py
```diff
@@ -0,0 +1,398 @@
...
+            lab_st1 = np.array(['FN'] * len(sts1[u_i]))
+            # Sam note : I would add this line
+            #�labels_st1[u1] = lab_st1
```
alejoe91: yes please add

spiketoolkit/comparison/comparisontools.py
```diff
@@ -0,0 +1,398 @@
...
+        st1 = sts1[u_i]
+        for l_gt, lab in enumerate(lab_st1):
+            if lab == 'UNPAIRED':
+                for u_j, u2 in enumerate(sorting2.getUnitIds()):
+                    if u2 in unit_map12.values() and unit_map12[u1] != -1:
+                        lab_st2 = labels_st2[u2]
+                        st2 = sts2[u_j]
+
+                        n_up = st1[l_gt]
+                        id_sp = np.where((st2.astype(int)-n_sp)<=delta_tp//2)[0]
```
alejoe91: can you change these lines to:
```
matches = (np.abs(mapped_st.astype(int)-n_sp)<=delta_tp//2) 
if np.sum(matches) > 0:
    lab_st1[l_gt] = 'CL_' + str(u1) + '_' + str(u2)
    lab_st2[np.where(matches)[0][0]] = 'CL_' + str(u2) + '_' + str(u1)
```

spiketoolkit/comparison/comparisontools.py
```diff
@@ -0,0 +1,398 @@
...
+        u2 = unit_map12[u1]
+        if u2 !=-1:
+            lab_st1 = labels_st1[u1]
+            lab_st2 = labels_st2[u2]
+            mapped_st = sorting2.getUnitSpikeTrain(u2)
+            # from gtst: TP, TPO, TPSO, FN, FNO, FNSO
+            for sp_i, n_sp in enumerate(sts1[u_i]):
+                matches = (np.abs(mapped_st.astype(int)-n_sp)<=delta_tp//2)
+                if np.sum(matches) > 0:
+                        lab_st1[sp_i] = 'TP'
```
alejoe91: extra tab

spiketoolkit/comparison/comparisontools.py
```diff
@@ -0,0 +1,398 @@
...
+    unit_map12: dict
+        Dict of matching from sorting1 to sorting2.
+        
+
+    Output
+    ----------
+    
+    confusion_matrix: the confusion matrix
+    
+    st1_idxs: ???
```
alejoe91: st1_idxs: order of units1 in confusion matrix

st2_idxs: order of units2 in confusion matrix

spiketoolkit/comparison/sortingcomparison.py
```diff
@@ -205,280 +216,27 @@ def plotConfusionMatrix(self, xlabel=None, ylabel=None):
...
-
-        self._confusion_matrix = conf_matrix
-        st2_idxs = np.append(st2_matched, st2_unmatched)
-
+        self._confusion_matrix,  st1_idxs, st2_idxs = do_confusion_matrix(self._sorting1, self._sorting2,
+                                                self._unit_map12, self._labels_st1, self._labels_st2)
+        
         return st1_idxs, st2_idxs
 
+
```
alejoe91: Yes I guess this can be moved to comparisontools.

It performs a 1-to-1 spike train match

samuelgarcia:
@alejoe91 thanks !!!

spiketoolkit/comparison/sortingcomparison.py
```diff
@@ -534,20 +292,10 @@ def getUnitSpikeTrain(self, unit_id, start_frame=None, end_frame=None):
...
-    if (len(inds) == 0):
-        return 0
-    inds2 = np.where(inds[:-1] + 1 != inds[1:])[0]
-    return len(inds2) + 1
-
-
+# Samuel EDIT
+# I really do not andderstand what there is this function ???
+# is it another version of the confisnion matrix ?
+# is it done with another object model ?
```
alejoe91: yes this can be removed, it's a leftover

alejoe91:
hei @samuelgarcia, I requested some changes. I think after that we can merge :)

samuelgarcia:
@alejoe91 : I made the change you request but now tests are not working anymore see travis.

In test_compraisontools/test_do_counting()
the toy example change the FN to CL. Is it normal ?
In this example the latest spike is not detected (opposed to .not wweel assigned).
When a spike is not detected should it be a FN ? What is the difference between FN and CL ?

Could you help me ?

samuelgarcia:
And there are still some ```#Edit samuel``` in sortingcomparison.py to be anwsered.

spiketoolkit/comparison/comparisontools.py
```diff
@@ -0,0 +1,396 @@
...
+        lab_st1 = labels_st1[u1]
+        #~ st1 = sorting1.getUnitSpikeTrain(u1)
+        st1 = sts1[u_i]
+        for l_gt, lab in enumerate(lab_st1):
+            if lab == 'UNPAIRED':
+                for u_j, u2 in enumerate(sorting2.getUnitIds()):
+                    if u2 in unit_map12.values() and unit_map12[u1] != -1:
+                        lab_st2 = labels_st2[u2]
+                        st2 = sts2[u_j]
+                        matches = (np.abs(mapped_st.astype(int)-n_sp)<=delta_tp//2)
```
alejoe91: my bad: mapped_st is not changed from the previous loop.
I guess this should be fixed by adding after line 283:
```
n_sp = st1[l_gt]
mapped_st = sts2[u_j] # instead of st2 = sts2[u_j]
```
this should fix the CL problem.

spiketoolkit/comparison/sortingcomparison.py
```diff
@@ -23,9 +36,11 @@ def __init__(self, sorting1, sorting2, sorting1_name=None, sorting2_name=None, d
             self._do_counting(verbose=verbose)
 
     def getSorting1(self):
+        # Samuel EDIT : why not a direct attribute acees  with self.sorting1 ?
```
alejoe91: yeas I guess we could allow to access sorting1 ad sorting2 directly

spiketoolkit/comparison/sortingcomparison.py
```diff
@@ -23,9 +36,11 @@ def __init__(self, sorting1, sorting2, sorting1_name=None, sorting2_name=None, d
...
         return self._sorting1
 
     def getSorting2(self):
+        # Samuel EDIT : why not a direct attribute acees  with self.sorting2 ?
```
alejoe91: same here

spiketoolkit/comparison/sortingcomparison.py
```diff
@@ -153,6 +166,8 @@ def computeCounts(self):
             self._do_counting(verbose=False)
 
     def plotConfusionMatrix(self, xlabel=None, ylabel=None):
+        # Samuel EDIT
+        # This must be moved in spikewidget
```
alejoe91: yes you can remove this. We can move it later from previous commits

spiketoolkit/comparison/comparisontools.py
samuelgarcia: I think I made this change at the wring place. No ?

spiketoolkit/comparison/sortingcomparison.py
```diff
@@ -41,9 +56,11 @@ def getLabels2(self, unit_id):
             raise Exception("Unit_id is not a valid unit")
 
     def getMappedSorting1(self):
+        # Samuel EDIT : the use case of this must documented
```
alejoe91:
```
Returns a MappedSortingExtractor for sorting 1. The returned MappedSortingExtractor.getUnitIds returns the unit_ids of sorting 1. The returned MappedSortingExtractor.getMappedUnitIds returns the mapped unit_ids of sorting 2 to the units of sorting 1 (if units are not mapped they are labeled as -1). The returned MappedSortingExtractor.getUnitspikeTrains returns the the spike trains of sorting 2 mapped to the unit_ids of sorting 1.  
```

alejoe91:
@samuelgarcia the suggested change should fix the problem. The problem was that mapped_st was not updated from the previous loop.

spiketoolkit/comparison/sortingcomparison.py
samuelgarcia: So are you OK for _sorting1 to sorting1 and remove this ?

spiketoolkit/comparison/sortingcomparison.py
samuelgarcia: Before remove it do you want to copy/paste in spikewidget ?
Maybe wait to refactor spikewidget before ?

samuelgarcia:
@alejoe91 
Thank you for the help.

Are you agree to remove the list of spiketrains introduced by Matthias and keep only dict for spiketrains ?
This double indexes (key + number) is hard to read. No ?

colehurwitz31:
@samuelgarcia Do you want me to put everything into snake case on this branch? That would solve a lot of travis issues.

alejoe91:
> @alejoe91
> Thank you for the help.
> 
> Are you agree to remove the list of spiketrains introduced by Matthias and keep only dict for spiketrains ?
> This double indexes (key + number) is hard to read. No ?

Yes I agree :)

samuelgarcia:
@colehurwitz31 : thank you for the proposal. I think the fix about the case, I think this should be done outside this PR. You can start in parralel.
I will force the version of spikeextractors in travis. Or we can merge this even if test do not pass on travis if they pass on my mahcine (with no upgrade in spikeextractors).

@alejoe91 : I will do tomorrow (today?) a very last commit about the dict/list in comparisontools and then we will be able to merge.

colehurwitz31:
Ok, once you merge, I will refactor spiketoolkit to be up-to-date with the latest spikeextractors.

alejoe91:
Ok, we can do the refractoring once it's merged :)

samuelgarcia:
I made some other change in comparisontools.
You can merge now.

I will go back on this next week.

@colehurwitz31 : after the big change for case don't forget to change the travis file:
  * remove line 20
  * uncomment line 19
So the travis go back to spikeextractor master.

PR #74, opened 2019-03-19
First example of post processing module based off of CuratedSorting

colehurwitz31:
Hoping to get some feedback on this class. I wanted it to be similar to the filterrecording structure where each postprocessing module is a class based off of the CuratedSortingExtractor.

@alejoe91 @samuelgarcia

### DIFF
--- a/examples/sorters_example.ipynb
+++ b/examples/sorters_example.ipynb
@@ -115,16 +115,7 @@
    "outputs": [],
    "source": [
     "# Tridesclous\n",
-    "sortin_TDC = st.sorters.run_tridesclous(recording, output_folder='tmp_TDC')\n"
-   ]
-  },
-  {
-   "cell_type": "code",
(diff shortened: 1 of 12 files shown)


alejoe91:
Hi Cole!

It looks good to me but I would change the naming as I think it can be a bit confusing.

colehurwitz31:
I want to have threshold in the name so that they know that it will get rid of units below that threshold. I can think about a better name that still contains threshold though

colehurwitz31:
How does that look?

colehurwitz31:
Got a weird travis error:

KeyError: "Unable to open object (object 'clusters' doesn't exist)"

In the Klusta test suite... Have you seen this before?

@alejoe91 @samuelgarcia

colehurwitz31:
Ready to merge when you are @alejoe91

PR #80, opened 2019-03-28
Exposed all objects for GUI

colehurwitz31:
Exposed all objects for GUI

### DIFF
--- a/spiketoolkit/postprocessing/__init__.py
+++ b/spiketoolkit/postprocessing/__init__.py
@@ -1,2 +1,2 @@
 from .postprocessing import getUnitWaveforms, getUnitTemplate, getUnitMaxChannel, computePCAScores,  exportToPhy
-from .threshold_min_num_spikes import threshold_min_num_spikes
+from .threshold_min_num_spikes import threshold_min_num_spikes, ThresholdMinNumSpike
--- a/spiketoolkit/preprocessing/__init__.py
+++ b/spiketoolkit/preprocessing/__init__.py
@@ -1,7 +1,7 @@
-from .bandpass_filter import bandpass_filter
-from .notch_filter import notch_filter
-from .whiten import whiten
-from .common_reference import common_reference
-from .resample import resample
-from .rectify import rectify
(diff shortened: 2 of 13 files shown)


colehurwitz31:
This allows the GUI to get the preprocessing classes from spiketoolkit rather than just the functions. Also, I removed some asterisks from preprocessing that forced keywords.. they were annoying. If you want me to put them back, then I will, but I think they are unnecessary.

colehurwitz31:
It is ready to be merged whenever

samuelgarcia:
Hi @colehurwitz31 .
If I anderstand correctly you force keywaord argument anywhere in this PR to make the GUI easier.

Don't you plan to in add some attribute directly in tookit to describe gui_params ?

Like:

```python
class BandpassFilterRecording:
  _gui_parms = [
        {'name':'freq_min', 'value': 300., 'type':'float'},
        {'name':'freq_max', 'value': 6000., 'type':'float'},
    ...
    ]
    def __init__(...):
         ...
```

colehurwitz31:
I don’t actually force keyword arguments anywhere. I removed the asterisks that force keyword arguments. The only changes I made here other than that is exposing the objects.

Yes, I plan to add GUI params soon, but I ran into a bug with the the PyQtGraph Parameter library that I was trying to solve first where default values are not working as intended. I am also planning on implementing extractor_list, preprocessing_list, and posprocessing_list similar to sortinglist.

samuelgarcia:
OK sorry.
I read this too fastly.

samuelgarcia:
This is OK for me.

My only concern is that you made some pep8 modification on launcher and BaseSorter since I am working on this in #73 I hope that I won't have conflict.

colehurwitz31:
I actually don’t think I really changed anything in those two modules (all the spaces I deleted, I added back) So I don’t think there will be any conflicts, but I am not sure. I probably shouldn’t have added them to the commit.

samuelgarcia:
No worry, we will see git shoudl be able to merge this :)

PR #82, opened 2019-04-03
Avoid copy the entire when it is a BinDatRecordingExtractor with framefirst=True

samuelgarcia:
see #71 
Several sorter naturaly work with the BinDatRecordingExtractor as input:
  * tridesclous
  * kilosort
  * kilosrt2
  * klusta
  * circus (with some modification)
For the moment, the recording is always copied into private file even if the input is already the correct format.

In this PR we check if the input recording has already the correct format and in that case there is no copy at all.

This should save ressource with the launcher because recording could be copied only once for all sorter making the comparison much fatser with less disk footprint.

@colehurwitz31 @alejoe91 : I did it only for tridesclous at the moment. If you validate this I will propagate this to others sorters.


colehurwitz31:
Yes, I think this is good! The other sorters should have something similar.

alejoe91:
@samuelgarcia good idea. 
Also I think we should make the changes that we made to the writeBinaryFormat (time_axis etc) into the write_recording static method of the bindat extractor and use that instead.

What do you think?

samuelgarcia:
@alejoe91 I think it is good to have a simple and indepedant function in tools.py for that.
I think we just should reimpleement BinDatRecordingExtractor.writeRecording that would be a call to this function. No ?
For the moment until Cole have finish the big change about the case, I still have an old version of spikeextractor. I prefer to not make any change now in spikeextractor and focus only on spiketoolkit.

samuelgarcia:
This PR is quite old and I won't do much more on this.

Basically it make possible to not copy inside the path the recording in case of BinDatRecordingExtractor for tridesclous an klusta.
It also add the possibility in the launcher to copy everything inside and share file across sorter. This avoid many copy.

If travis pass test, then ready to merge.

samuelgarcia:
Travis is failing because of HS.
I think a new change in HS must trigger a new bug here.
This have nothing todo with this PR.
I think we can merge. So I could go back to comparison and improve the **compute_performance** with by cluster metrics.

alejoe91:
Ok merging :)

alejoe91:
merged

PR #83, opened 2019-04-03
Revenge of the Snake Case

colehurwitz31:


colehurwitz31:
Close to being refactored! The only failing test is the mountainsort test because I think it is locally using an older version of spikeextractor. @magland Can you update ms4 to work with spikeextractors 0.4.0?

samuelgarcia:
Amazing work.
Waouh.
the travis is more complaining about a "getChannelIds". Is it the problem ?

alejoe91:
@samuelgarcia yes ms4 uses internally spikeextractors, so we need to update it. I can try to do it tomorrow in my fork and send a PR to @magland

samuelgarcia:
OK.
I think they should do a tag/release in MS4 before the case transition and then include the patch.

PR #94, opened 2019-04-24
Added docstrings

alejoe91:
- preprocessing
- postprocessing
- validation

will finish comparison before merging


alejoe91:
also removed curation folder

colehurwitz31:
Looks great! This will be super helpful for the GUI and for using these tools.

Could we rename q --> quality_factor? I think that would be a good change.

alejoe91:
in scipy it's called q, but I guess being clearer won't harm :)

colehurwitz31:
Hmm... I guess it may not be a big deal then. Up to you :)

alejoe91:
I will merge this and we can improve docstrings later on :)

PR #95, opened 2019-04-24
Parallel spike matching in do_matching() for speed

mhhennig:
Please check if this implementation is ok. If so, I will add something similar in other places.


mhhennig:
If we use this, we need joblib as requirement (Travis fails at the moment).

colehurwitz31:
Could you make n_jobs and parallel parameters of the sorting comparison? Then we can make it optional and then we can give an error message if they don't have joblib installed (like how we handle optional packages elsewhere)

samuelgarcia:
This is cool.
I think we can have a hard dependency on joblib. It is very light and all people have it installed yet (because of sklearn).

the solution of cole is also valid. Make a weak dependency.
And if joblib is not installed make a standard loop.

Could you add some test in test_comparisontools ?

mhhennig:
Sklearn has its own joblib, but I agree we should make it a hard dependency as it's useful in many contexts. Could you add this?
I will run more tests, let's leave this pull request open for now.

mhhennig:
Should this read
```
                    sp1 = self._sorting_list[self._name_list.index(sorter1)].get_unit_spike_train(unit1)
                    sp2 = self._sorting_list[self._name_list.index(sorter2)].get_unit_spike_train(unit2)
                    lab1, lab2 = SortingComparison.compare_spike_strains(sp1, sp2)
```
?

samuelgarcia:
Oups. My bad.
I remove this "compareSpikeTrains"  in the refactoring PR #73.
See discussion with @alejoe91 about that
I made big clean because some method were redundant and not used anywhere and not documented.
I am really sorry this function need to be integrated again.
I propose to put as a function with proper and independant unitest in comparisontools.

samuelgarcia:
I think we should also make a test_multisortingcomparison.py to avoid this in the futur when a refactoring occur. I am sorry.

mhhennig:
No worries - would you mind adding it in this branch? Once that's working again, I will test more and see if more parallel code is useful.

colehurwitz31:
Added the joblib dependency. It only fails now because of a herding spike thing (can't find positions_spikeextractor)

samuelgarcia:
@mhhennig : I will try to fix this morning and add a test_multisortingcomparison.py.

```compare_spike_strains ``` is missing but I have also a bug in this code:
```
  File "/home/samuel/Documents/projet/SpikeInterface/spiketoolkit/spiketoolkit/comparison/multisortingcomparison.py", line 18, in __init__
    self._do_matching()
  File "/home/samuel/Documents/projet/SpikeInterface/spiketoolkit/spiketoolkit/comparison/multisortingcomparison.py", line 112, in _do_matching
    sp1 = self._sorting_list[self._name_list.index(sorter1)].get_unit_spike_train(unit1)
ValueError: sequence.index(x): x not in sequence
```

This is because keys form _name_list are integer and not str. I will also do this change.

**Comment by samuelgarcia** [comment]
@mhhennig @colehurwitz31 @magland @alejoe91 : I have implemented back the "compare_spike_strains". I made it as a function rathed than a static method. This make the architecture easier.

Please have a look carrefully in the implementation : I think this is incorrect. See the ```test_compare_spike_trains()``` in test_comparisontools.py.
The two mistakes in my mind:
  * The way it is implemented is bassed on the fact that the spiketrain1 is the groung truth. 
     So this implemention should not be used in MultisortingComparison where the GT is unkown.
  * When 1 spike from spiketrainA match 2 spikes in spiketrainB then they are counted as "FP" both.
    This is wrong, one should TP and one should FP.

**Comment by samuelgarcia** [comment]
Also: some explicit code in tests should be add for the joblib option.

**Comment by alejoe91** [comment]
@samuelgarcia @mhhennig  Should we keep this open and fix those 2 mistakes before merging?

I won't have time before monday, but I can prioritize it then.
I'll also complete the docstrings for all functions.

**Comment by samuelgarcia** [comment]
No worry.
I did not explore totally the MultiSortingComparison and so don't known the consequence of this.

Note that in do_counting we have:
```python
                matches = (np.abs(mapped_st.astype(int)-n_sp)<=delta_tp//2)
                if np.sum(matches) > 0:
                    lab_st1[sp_i] = 'TP'
                    lab_st2[np.where(matches)[0][0]] = 'TP'
```
which is more correct!

The best would to use internally the compare_spike_train in do_counting to avoid duplicated code that have a big impact on the result.

alejoe91:
@mhhennig this is now merged with @samuelgarcia changes. I'll make some more tests and merge this too.

alejoe91:
@samuelgarcia why there is a `false_disc_rate` and a `false_discovery_rate` in the dataframe from `get_performance`?

samuelgarcia:
I don't known this is maybe a bug or an inconsistensy. I changed some names to explicitly make "rate".

I also want to be able to export "raw" counting datafarme not in ratio.

samuelgarcia:
@alejoe91 
did you fix this in MultiSortingComparison which is not a bug but a mistake ?
```python
                matches = (np.abs(mapped_st.astype(int)-n_sp)<=delta_tp//2)
                if np.sum(matches) > 0:
                    lab_st1[sp_i] = 'TP'
                    lab_st2[np.where(matches)[0][0]] = 'TP'
```

alejoe91:
I'm changing both to `false_discovery_rate`. There was also a misordering in get_performance pulled_by_sum. I'll push soon and you can take a look

alejoe91:
> @alejoe91
> did you fix this in MultiSortingComparison which is not a bug but a mistake ?
> 
> ```python
>                 matches = (np.abs(mapped_st.astype(int)-n_sp)<=delta_tp//2)
>                 if np.sum(matches) > 0:
>                     lab_st1[sp_i] = 'TP'
>                     lab_st2[np.where(matches)[0][0]] = 'TP'
> ```

No, where should I change this?

samuelgarcia:
Oups. My bad this is the correct version.
I don't remember where but there is a 
```if np.sum(matches) == 1:```
in the code and this is incorrect.

I will have a look.

samuelgarcia:
This is in compare_spike_train added in this PR directly.

samuelgarcia:
her https://github.com/SpikeInterface/spiketoolkit/pull/95/commits/b6f2bfacacbebcfb8d70f6edaf7c55a29bbd146c
line 436
```if len(id_sp) == 1:```

alejoe91:
ok fixed :)

Debugging the multisortingcomparison now. I found a small bug in the agreement score

alejoe91:
@mhhennig @samuelgarcia I made some changes in the sortingcomparison and multisortingcomparison:

- now they are called with: `compare_two_sorters` and `compare_multiple_sorters` (in line with other function-based modules)
- I went through the multicomp once again and fixed it. Added a test. @mhhennig where is the code to plot the network? I want to move the plotting to spikewidgets
- added docstings to comparison functions

I think we can merge and be satisfied with the comparison module :)

alejoe91:
> @mhhennig @colehurwitz31 @magland @alejoe91 : I have implemented back the "compare_spike_strains". I made it as a function rathed than a static method. This make the architecture easier.
> 
> Please have a look carrefully in the implementation : I think this is incorrect. See the `test_compare_spike_trains()` in test_comparisontools.py.
> The two mistakes in my mind:
> 
> * The way it is implemented is bassed on the fact that the spiketrain1 is the groung truth.
>   So this implemention should not be used in MultisortingComparison where the GT is unkown.
> * When 1 spike from spiketrainA match 2 spikes in spiketrainB then they are counted as "FP" both.
>   This is wrong, one should TP and one should FP.

@samuelgarcia the multisortingcomparison only uses the agreement fraction and reconstruct a sorting extractor only with TP spikes. I think it's safe to use it.
This should make spike trains much cleaner!

alejoe91:
> Should this read
> 
> ```
>                     sp1 = self._sorting_list[self._name_list.index(sorter1)].get_unit_spike_train(unit1)
>                     sp2 = self._sorting_list[self._name_list.index(sorter2)].get_unit_spike_train(unit2)
>                     lab1, lab2 = SortingComparison.compare_spike_strains(sp1, sp2)
> ```
> 
> ?

Yes, fixed all :)

alejoe91:
Can I go ahead and merge? @mhhennig @colehurwitz31 @samuelgarcia

PR #98, opened 2019-05-02
fixed bug in klusta and worked on notebooks module by module

alejoe91:
- bug in klusta fixed (wrong dtype in config file)

- organized all notebooks by module and made them nicer. Still working on comparison and validation one


samuelgarcia:
I have not checked all notebook but the "comparison_example" don't have the figure rendered.
I think this is nicer to have figures and results rather than empty cell because it help the reader to understand the code without running the NB. No ?

colehurwitz31:
We can leave the figured rendered, but I think that increases the github history which makes cloning it take longer? I can't remember if that is true or not though...

alejoe91:
One thing we can do is export the notebooks with figures to rst to build quickly the read the docs.
At this point I recommend having a read the docs for each repo, so it' easier to show the API.

For spikeextractors we can have:
- RecordingExtractor
- SortingExtractor
- Available extractors
- Sub and Multi Extractors
- Dealing with probe files

For spiketoolkit we can have it by module:
- Preprocessing
- Sorters
- Postprocessing
- Comparison
- Validation

For spikewidgets we can discuss about it.

@samuelgarcia I'll go through your PR first and if all is clear I'll merge. Then I'll do Matthias'. 
Finally I'll finish this one and pull and modify the sorter_comparison one from the merged branch :)

samuelgarcia:
There is sphinx plugin galery wich is cool because you write a script and it is converted to both sphinx page and downloadable notebook.
They use this everywhere.

With this example do not have figures, because they simple script. But when we build sphinx figure are done wihtin notebooks. It avoid also git noise dur to notebook inside.
i think we should go for this.

https://github.com/sphinx-gallery/sphinx-gallery

alejoe91:
great idea! If you used it before can you start adding it here?

alejoe91:
Ok with this PR:

- a klusta bug is fixed
- notebooks are nice and clean (the comparison notebook is awesome @samuelgarcia : I added an initial part and left yours as 'Systematic performance comparison on ground truth'
- further debuggin of multisorting comparison, with tests and explanatory example in notebook.
- implenented a `remove_bad_channels` preprocessing function. It's more straightforward than manually doing the SubRecordingExtractor and it also has an automatic mode

I left the notebooks without plots, but we can discuss about the sphinx-gallery on monday.

alejoe91:
write test for multisorting and reverse order

alejoe91:
merging this :)

PR #104, opened 2019-05-07
fix export to phy

colehurwitz31:


alejoe91:
Did you find errors in the other PR? Otherwise that's the intended behavior and you can close this :)

colehurwitz31:
I found an error with it. It was saying that templates is a list so it doesn't have shape

colehurwitz31:
So the notebook kept failing

alejoe91:
Ok, but that's not a fix ;)

Try to substitute the if with:

if not isinstance(templates, list):

If more units, templates should be a list

colehurwitz31:
Hmm, I guess I am not sure what templates is... I thought it was just a list. What else can it be?

alejoe91:
Array

PR #111, opened 2019-05-09
Add new performence table : count nb_units above threshold

samuelgarcia:
This count how many units above threshold (95% by default) have been detected.
This is done on ['tp_rate', 'accuracy', 'sensitivity'].

Spikeforest have a similar metric but not implemented in spiketoolkit.
It is important to have it.

### DIFF
--- a/spiketoolkit/comparison/gathersortingcomparison.py
+++ b/spiketoolkit/comparison/gathersortingcomparison.py
@@ -6,7 +6,7 @@ import pandas as pd
 from spiketoolkit.sorters import run_sorters, collect_results
 from .sortingcomparison import compare_two_sorters, _perf_keys
 
-def gather_sorting_comparison(working_folder, ground_truths, use_multi_index=True):
+def gather_sorting_comparison(working_folder, ground_truths, use_multi_index=True, threshold=95):
     """
     Loop over output folder in a tree to collect sorting from
     several sorter on several dataset and returns sythetic DataFrame with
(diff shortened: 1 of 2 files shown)


spiketoolkit/comparison/gathersortingcomparison.py
```diff
@@ -6,7 +6,7 @@
 from spiketoolkit.sorters import run_sorters, collect_results
 from .sortingcomparison import compare_two_sorters, _perf_keys
 
-def gather_sorting_comparison(working_folder, ground_truths, use_multi_index=True):
+def gather_sorting_comparison(working_folder, ground_truths, use_multi_index=True, threshold=95):
```
colehurwitz31: Can we call this accuracy_threshold or acc_threshold? Otherwise, it is a bit confusing what it is thresholding.

colehurwitz31: Also, does 95 seem a bit high? Maybe 90 or something? Not sure we want a threshold that high as default? I don't really know though

colehurwitz31:
A great addition. I have been wanting this :)

samuelgarcia:
Have a look also here https://github.com/SpikeInterface/spikeresults/pull/2.
Latest figures use this metric.

samuelgarcia:
In that directon we shoulc also have count under threshold to detect how many cell are not detected at all.

samuelgarcia:
If all of you are happy with this we can merge.
I will come back with performence metric after discussion.

alejoe91:
Yep merging :)

PR #112, opened 2019-05-09
various fixes fixes to export to phy (now dealing with groups), kilosort1/2 (car as…

alejoe91:
fixes to export to phy (now dealing with groups), kilosort1/2 (car as param), launcher


spiketoolkit/sorters/basesorter.py
```diff
@@ -58,10 +58,10 @@ def __init__(self, recording=None, output_folder=None, debug=False,
             # several groups
             self.recording_list = se.get_sub_extractors_by_property(recording, grouping_property)
             n_group = len(self.recording_list)
-            if n_group>1:
-                self.output_folders = [Path(str(output_folder) + '_'+str(i)) for i in range(n_group)]
+            if n_group > 1:
+                self.output_folders = [output_folder / str(i) for i in range(n_group)]
             else:
-                self.output_folders = [output_folder]
+                self.output_folders = [output_folder / str(0)]
```
samuelgarcia: Does it break launcher ?

alejoe91: Seems to pass the tests. I think we if the gather_comparisons.. can deal with recordings sorted by group. Let's keep this open, I will test tomorrow

samuelgarcia:
@alejoe91 I need to make some test in launcher and in gather_result with this changes.

I have a small issue with launcher / gather machinery because it relly on the fact that you compare one recording that will not be splitten otherwise it is a total mess for final result (some recording could have several output line wheras other would have only one line).

So I need to go back in the code and see the consequence. Unfortunatly I won't have time until next Tuesday...

alejoe91:
I can try to fix it :) should be easy with multisorting extractor

samuelgarcia:
The wird thing is the following : 

I designed the lancher gather result in a simple way in case of unique recording.
The main reson is that when we have a ground truth linked to a recording. In that case the sortingextractor (GT) is related to ONE recording and not a recording that must be split in several sub recording.
So in that case mixing subrecording and multisorting extractor lead to over tricky gather result.

Having the ablity to split a recording on something (= channel group) in case of multi tetrode is a really cool feature to have.

But this same feature is not cool in case of (multiple sorter X multiple recording) comparison with launcher/gather_result because each recording is implicitly linked to one relative recording.

Have a look the notebook of comparison I send yesterday and imagine you have several recordings.
what would be the result output if one of the recording should be split in many tetorde.
The final result would be a nested result and not a simple table.

So I think, we have to forbid the case of splitable recording in the pipeline launcher>gather_result.
If a recording have to be splited then it have to be splitted before the chain explicitly and not inside each sorter vith the SorterBase trick.

samuelgarcia:
Neverless I think, it is good to have
``` self.output_folders = [output_folder / str(0)]``` even for none splitable recording because it make general multi segment even if one segment.

So basically you can merge, but I will need to come back on launcher>gather_result chain later on anyway.

**Comment by alejoe91** [comment]
I think it's as easy as:

```
sorting_list = []
for folder in output_folders:
    sorting_list.append(CorrectSortingExtractor(folder))
sorting = se.MultiSortingExtractor(sortings=sorting_list)
```

this would work in both cases (that's why I made an extra folder 0) in case there are no groups.

What do you think?

alejoe91:
I'll merge in the meanwhile :)

PR #114, opened 2019-05-10
Now saves the sampling rate, also minor default parameter adjustments

mhhennig:
Some changes to the HS2 sorter class.


samuelgarcia:
Some people are fighting for benchmark  in  that zone :)

PR #119, opened 2019-05-14
scale_traces --> transform_traces

colehurwitz31:
Traces will be scaled by: traces*scalar + offset in this preprocessing tool. Is this what you had in mind Matthias?


colehurwitz31:
Default values of: scalar = 1 and offset=0

Issue #113, opened 2019-05-09
validation metrics do not agree


mhhennig:
I compared the results of a `sortingcomparison`, and they are inconsistent.

Example from the table returned by `sc.get_performance('by_spiketrain')`:

unit | tp_rate |	fn_rate | cl_rate | fp_rate_st1 | fp_rate_st2 | accuracy | sensitivity| precision | miss_rate | false_discovery_rate
-|-|-|-|-|-|-|-|-|-|-
0|0.516944|3.67605|0.516944|388.742|94.9095|0.13156|12.3288|0.132802|87.6712|99.8672

Result from `sc.get_false_negative_fraction(0)` and `sc.get_false_positive_fraction(0)`, and I get:

0.15623205054566336  0.5799256505576208

This looks inconsistent, is there a difference in how these are computed?

alejoe91:
I haven't checked the single functions lately. I think Jeremy wrote those in the first place. I think we should remove them and land on the get_performance call :)

I would trust the get_performance ones, but it's good to double check

mhhennig:
Ok, agree, that's good.
Does `sc.get_best_unit_match1(unit)` still do the right thing, i.e. return the one that has been established as the best match?

alejoe91:
Yes the matching should be the same

samuelgarcia:
I agree with @alejoe91 

If @magland wrote ```get_false_negative_fractio``` and ```get_false_positive_fraction``` , I am afraid that they are used in the spikeforest website.
I think theses metrics are bias because the fraction is done on len(units2) and should done on len(units1) like in get_performance().

I think that get_performance should also have a mode to return "raw_count" in a DataFrame with no fraction at all.

And if th issueis open, I would propose to remove percent scale (0-100) and replace by ratio scale (0-1).
We can keep 100 factor on print but not internally.
This percent scale will lead us to make mistake I guess.
Would you be OK to go the fraction ?

alejoe91:
I'm ok with this

samuelgarcia:
I will patch get_false_negative_fractio and get_false_positive_fraction and use internally get_performance.

This will avoid duplicate code that are not agree.

In the same PR I plan:
  * to go for scale [0-1]
  * export the dataframe of raw_count
  * count how many nott match units


OK ?

samuelgarcia:
I am reading again the entire code of SortingComparison.

What get_labels2 is for ? There is also something wrong in it.

[...]

PR #121, opened 2019-05-15
Refactoring SortingComparison

samuelgarcia:
See #113 
Refactoring SortingComparison split into SortingComparison + GroundTruthComparison


samuelgarcia:
@alejoe91 @colehurwitz31 : I have started the refactoring.

Could you have a look.

There is also now a GroundTruthComparison.get_raw_count that convert the internal counts dict into a dataframe. Could you tell what do you think.
I think we could also reformulate performance  'pooled_with_sum' and 'pooled_with_average' based on this table.

I also propose but this must be discussed : 
  * remove  'fp_rate_st2' and keep only  'fp_rate_st1' renamed in into  'fp_rate'

Now that there is a clear separation of teh goal of the class this columns have no inmeaning.
What do you think ?

colehurwitz31:
Good changes! I agree with getting ride of fp_rate_st2 for ground truth. Also, the raw count table would be helpful to have.

samuelgarcia:
OK I stop for today.
more change tomoroow.

samuelgarcia:
@colehurwitz31 @alejoe91 : are we agree that the actual do_couting is valid only in the case GroundTruthComparison ?
So I propose to also remove countin from SortingComparison.
We will be able add other "counting" in SortingComparison if necessary but the catual one is ground truth oriented one.

samuelgarcia:
I propose to also remove the count=True/False option at instanciate step because it is very fast and must be done always. It add useless kargs in function.

colehurwitz31:
Agreed on removing the count parameter. It causes more issues for me than it is worth.

samuelgarcia:
@alejoe91 @colehurwitz31 
count args is no mroe here and the counting logic is now only in GroundTruthComparison.
Could you check this ?

samuelgarcia:
I now plan remove the **do_counting()** from comparisontools.
The same logic will be implemented in GroundTruthComparison directly but based only on DataFrame.
The actual way nested list of dictionary, it can be now simplified with using only the get_raw_count table.

@colehurwitz31 @alejoe91 : are you agree with that ?

alejoe91:
It looks good to me @samuelgarcia

alejoe91:
Why is the gather_sorting_comparison test failling?

colehurwitz31:
sorting_comp = compare_sorter_to_ground_truth(gt_sorting, sorting, count=True)
E               TypeError: compare_sorter_to_ground_truth() got an unexpected keyword argument 'count'

Samuel removed the "count" keyword, but it is still in the test

samuelgarcia:
Thanks.
I saw, I will commti the fix with more changes.

samuelgarcia:
I have just push another big change.
All counting in GroundTruthComparison is now based on dataframe.
Easier to read that loop over dict (and maybe faster?)

Note that the new self.count is now the "raw" count without rate.

Rate are computed on the fly depending on the method.

samuelgarcia:
I also want to change the **get_number_units_above_threshold** and have a **get_correct_units** with differnts threshold combinied like ```(fp_rate>0.95) & (fn_rate)<0.05)``` this would give how many correct units

colehurwitz31:
I think you mean (tp_rate>0.95), but sounds good. Again, I would think above 90 is probably best

samuelgarcia:
This totally depend:
  * on the dataset quality
  * your own faith in spike sorting machinery
  * severity level

... and of course it will be a kargs of course :)
So chapel would be happy.

colehurwitz31:
All the changes look great so far btw :)

samuelgarcia:
Thank you.
I am quite happy because the GroundTruth part is now very clean and super easy to enhance.
The SorintgComparison zone still need some debugging... but it is not very used at the moment.

colehurwitz31:
I think we are planning to use SortingComparison for a lot of the upcoming paper so we probably want to make sure that is working as intended

samuelgarcia:
I think with the new design SpikeComparison is no more very usefull directly unless for MultiSortingComparison or MappedSortingExtractor.

There are still some error I think in get_false_positive_fraction, get_false_negative_fraction and we have to find a good deniminator.

Note that SortingComparison hinerits BaseComparison so it still have the same matching/score_labeling/confusion_matrix.
In the actual state:
 * **do_ matching** works in bijective way so no assumtion is done on wich is the GT.
 * **do_score_labels** is not symetrical because for CL it work only in one way. We have to think about 
    this. Maybe there is a bias, maybe not.
 * **do_confusion_matrix*** should be bijective, we have to check it in unitest this is not done.

So I don't known what we plan to put in the upcoming paper about this SorintgComparison (which is not used anymore for performance benchmark) We have to be very carefull about what we claim.

The GroundTruthComparison is  much easier for a clear description. There is no ambiguity.

samuelgarcia:
@alejoe91 @colehurwitz31 

See theses new methods:
get_well_detected_units/count_well_detected_units/get_fake_units_in_other/count_fake_units_in_other

If you are OK with this, I will propagate theses change to gather_comparison and example.

Then only bug in SortingComparison will be needed.

samuelgarcia:
OK done.

This PR is ready to reviewed.
In my point of GroundTruthComparison is OK unless bugs.

The old SortingComparison must be rework.
For instance as get_performance no longer exists there.

Notebook example comparison section **2) Compare multiple spike sorters** is no longer valid.
This is because there is a call to **get_performance**.
I think here we must rewrite some performance of agreement that are diffrents than previous one otherwise it is a theoritical mistake. I would also change the semantic of **SortingCOmparison.get_performance** to something else to not mix with **GroundTruthComparison.get_performance** for instance **SortingCOmparison.get_agreement_scores**

If you are OK with this PR, we could merge and the work for SortingCOmparison would be done later on collectively with more discussion.
I am not sure to be alignied with what you have in mind about MultiSortingComparison. Because MultiSortingComparison was using get_performance but performance was computed with GT in mind Wich is I think a mistake.

alejoe91:
I don't think that the MultiSortingComparison will be affected, as it only uses the `get_agreement_fraction`.
Maybe the get_performance is used in the notebook, but we could remove that.

I'll go through the PR today and get back :)

spiketoolkit/comparison/groundtruthcomparison.py
```diff
@@ -0,0 +1,291 @@
...
+        return perf[keep].index.tolist()
+    
+    def count_well_detected_units(self, **kargs):
+        """
+        Count how many well detected units.
+        Kargs are the same as get_well_detected_units.
+        """
+        return len(self.get_well_detected_units())
+    
+    def get_fake_units_in_other(self):
```
alejoe91: I would call it `unmatched_units`

spiketoolkit/comparison/groundtruthcomparison.py
```diff
@@ -0,0 +1,291 @@
...
+        The name of sorter 1
+    other_name: : str
+        The name of sorter 2
+    delta_frames: int
+        Number of frames to consider coincident spikes (default 10)
+    min_accuracy: float
+        Minimum agreement score to match units (default 0.5)
+    exhaustive_gt: bool (default True)
+        Tell if the ground true is "exhaustive" or not. In other world if the
+        GT have all possible units. It allows more performance measurment.
```
alejoe91: typo: measurement

spiketoolkit/comparison/groundtruthcomparison.py
```diff
@@ -0,0 +1,291 @@
...
+            self.count.loc[u1, 'nb_gt'] = self._labels_st1[u1].size
+            self.count.loc[u1, 'other_id'] = u2
+
+            if u2==-1:
+                self.count.loc[u1, 'fp'] = 0
+                self.count.loc[u1, 'nb_other'] = 0
+            else:
+                self.count.loc[u1, 'fp'] = np.sum(self._labels_st2[u2] == 'FP')
+                self.count.loc[u1, 'nb_other'] = self._labels_st2[u2].size
+
```
alejoe91: in the base sorting comparison I would add a `get_confusion_matrix()` function. This is useful for both the GroundTruthComparison and the SortingComparison and it will be usefool for the confusion matrix widget when we port it

alejoe91: this comment was meant for the BaseSortingComparison

spiketoolkit/comparison/basecomparison.py
```diff
@@ -0,0 +1,134 @@
+import spikeextractors as se
+
```
alejoe91: add `import numpy as np` for the moment to plot confusion matrix

spiketoolkit/comparison/groundtruthcomparison.py
samuelgarcia: I also don't like "fake_units".

But here we need a clear semantic. The pathology we are tracking here is that some sorter have many more units they should have and worse: sometime a gt units is matched to several units in sorter. The best one with a good TP score, the others matched must counted and should penalize the final benchmark. This is the goal of this function.

The semantic "unmatched_unit" in my opinion is more related to a not detected units, in short unmatched in the GT (the other way).

Could a good english speaker could make other other proposal ?
@colehurwitz31 @mhhennig ?

spiketoolkit/comparison/groundtruthcomparison.py
samuelgarcia: OK

spiketoolkit/comparison/basecomparison.py
samuelgarcia: OK

samuelgarcia:
@alejoe91 :I have done changes.
We still need to find a name for **get_fake_units_in_other** see my comments.

spiketoolkit/comparison/groundtruthcomparison.py
alejoe91: then they should be false_detected_units?

samuelgarcia: Quite good. lets wait other proposal.
The most important will be the documentation coming with this.

colehurwitz31: What about “false_units”? Or is that too vague?

colehurwitz31: Or maybe, “extra_units”

colehurwitz31: Or maybe “fake_units”

samuelgarcia: héhé fake_units was the proposal but Alessio do not like it. I am also not very happy with it.
I like when samtic about metric are cristal clear.

samuelgarcia:
@alejoe91 @colehurwitz31 @mhhennig 
please do nto merge soon, I have just realize that "get_fake_units_in_other" is underestimaing the fact.
Because there are 2 levels of fake_units:
  * the one implemnted, there is no match at all
  * a more insidious one : units in sorter2 that match GT but in a one way only, in short they are not the 
    best match of GT


This second pathology in shorter using template matching is really big. We nned something to track it.
In some sorter detect a the same GT units with several units in sorter2 and our match take the best one, the other should penalize the sorter in benchmark.

So we need 2 clear name for the method. Could you help me.
I will push a proposal of implementation I let you find a good naming.

mhhennig:
-> unmatched_units (can be in 1 and 2)?

samuelgarcia:
I agree that fake is not good.
Unmatched isnot precise enough.

To summarize the situation, we want to detect 2 pathology with 3 ensemble:
  1. "Unmatched at all"
```python
        fake_ids = []
        unit2_ids = self._sorting2.get_unit_ids()
        for u2 in unit2_ids:
            if self._best_match_units_21[u2] == -1:
                fake_ids.append(u2)
```
  2. **"not in best match GT"** = "matched but not best match" or "unmatched at all"
```python
        best_match = list(self._unit_map12.values())
        bad_ids = []
        unit2_ids = self._sorting2.get_unit_ids()
        for u2 in unit2_ids:
            if u2 not in best_match:
                bad_ids.append(u2)
        return bad_ids
```
  3. **"matched but not best match"** should be **2** without item of **1**
   It is not implemented yet but it is easy to do.

Now we have **3 names** to find.

In actual proposal, the naming is
1 = nb_fake
2 = nb_bad

alejoe91:
@samuel I think number 2 should be only "matched but not best match". This could be due to oversplits so I would call them either Oversplitted or Duplicates.

I'm ok for fake for 1. 

I don't really get number 3. Or maybe I got number 2 wrong and number 3 should be Oversplitted?

mhhennig:
I would not distinguish unit types here, this should be done by appropriate metrics.

We have a % agreement for each pair, and we have the best matched pairs. To inspect a result, I would then use these numbers to ask:
1. How well is each GT unit matched by the best matched sorted unit?
2. How many extra units (in addition to best matched units) do we have?
3. Are these extra units partial matches with a GT unit?

Essentially we ask if we over- or under-cluster the data, and whether detect noise as units. We should find the best metrics to quantify these.

colehurwitz31:
Would having an average_by_sum metric on Jeremy's website help alleviate some of the issues with how much noise the sorter makes?

samuelgarcia:
@alejoe91 : 
  * number 3 is "oversplitted only"
  *  number 2 is ``u2 not in list(self._unit_map12.values())```, so union of number 1 and number 3

samuelgarcia:
@mhhennig : I think we need to distinguish by type because before this PR they were not counted at all.
  * your number 1 was already done
  * your number 2 is my number 2 with other world
  * your number 3 is my number 3 with other world

samuelgarcia:
@colehurwitz31 does "average_by_sum" you mean "pool_by_sum" ?
What Jeremy do:
  * mode = "number of units above" is the new "count_well_detected_units" here
  * mode= "Average metric above SNR..." is "pool_by_average" but select only units obove SNR threshold

colehurwitz31:
Yes, I mean pool_by_sum. When we did this with kilosort, it did much worse.

mhhennig:
But were do you draw the line? What percentage counts as partial match versus completely different? In practise most units will have a little overlap, maybe a few percent, and some more.
Again, I think we should not focus too much on categorising units beyond the 'best match', but use complementary metrics instead. It may be worth looking at these in more detail:
https://scikit-learn.org/stable/modules/clustering.html#clustering-performance-evaluation

samuelgarcia:
@mhhennig : of course the line is abitrary (and user define). It is some kind of what is a good match from the user perspective.

In Jeremy website you can ajust it. the default is 0.8. I think it is too low.

In spikeinterface I put the default to 0.95 Cole was in favor of 0.9, you can change it in kargs anyway.

Avoid a treshold is always preferable, I agree with you.  If we find a global metric that cover everything, it would be the best. I will have a look in this doc.

The best trustable is the confusion matrix but as we need at some point to summarize many sorter for many dataset (the big picture), a threshold is needed, so one scalar to sumup the matrix...

spikeforest plot accuracy vs SNR but not include nor 1 nor 2 nor 3.

I think we could plot the ordered accuracy butit will be hard to represent many sorter and many dataset on one figure.

alejoe91:
I think it's nice to summarize with some metrics and also to differentiate between units. Maybe the well detected threshold should be 0.9.

Also @samuelgarcia can you change nb_... to n_... or num_...? 
I found nb a bit counterintuitive :)

samuelgarcia:
@alejoe91 : I will but what do I put after the underscore ? :)

alejoe91:
n_bad, n_fake etc :)

mhhennig:
Please not 'fake', it's not fake ;)

samuelgarcia:
I think Alesio was joking.
Any new idea for naming after a new night ?

alejoe91:
mmm...I think the most precise name would be `false_positive_units` for number 1. Number 2 can be `bad_units` (it's ok that it's more general), number 3 can be `oversplitted_units` or `partially_matched _units`

what do you think about this naming?

samuelgarcia:
OK for this.
If no other proposal I will do this soon.

samuelgarcia:
@alessio @cole @mhhennig  : I have just push the change for naming.

Please read careefully the **test_compare_sorter_to_ground_truth** test function as it summary possible errors (false_positive_units, oversplitted_units)

samuelgarcia:
I am wondering if "redundent_units" finnly would be better than "oversplitted_units" ?
I change another time.

samuelgarcia:
@alejoe91 @mhhennig @colehurwitz31 
Oups my first message have dispear.

I made the naming change with alesio proposal.

Maybe "redundent_units" finaly would be better than "oversplitted_units".

Please read carrefully **test_compare_sorter_to_ground_truth**.
Where all possible tracked mistake are tested.
Is is clear enough.

Please also validate the doc.

spiketoolkit/comparison/groundtruthcomparison.py
```diff
@@ -0,0 +1,392 @@
...
+            unit1_ids = self._sorting1.get_unit_ids()
+            perf = pd.DataFrame(index=unit1_ids, columns=_perf_keys)
+            
+            counts = self.count
+            perf['tp_rate'] = counts['tp'] / counts['num_gt']
+            perf['cl_rate'] = counts['cl'] / counts['num_gt']
+            perf['fn_rate'] = counts['fn'] / counts['num_gt']
+            perf['fp_rate'] = counts['fp'] / counts['num_gt']
+            
+            perf['accuracy'] = perf['tp_rate'] / (perf['tp_rate'] + perf['fn_rate']+perf['fp_rate'])
```
alejoe91: in our case `accuracy` and `tp_rate` is the same, no?

num_gt = num_tp + num_fn + num_fp --> 
accuracy = (num_tp / num_gt) / ((num_tp / num_gt) + (num_fn / num_gt) + (num_fp / num_gt)) = 
= num_tp / (num_tp + num_fp + num_fn) = num_tp / num_gt = tp_rate

@samuelgarcia @colehurwitz31 @mhhennig  right?

spiketoolkit/comparison/groundtruthcomparison.py
```diff
@@ -0,0 +1,392 @@
...
+        """
+        See get_bad_units
+        """
+        return len(self.get_bad_units())
+
+
+
+    
+# usefull also for gathercomparison
+_perf_keys = ['tp_rate', 'fn_rate', 'cl_rate','fp_rate',  'accuracy', 'sensitivity', 'precision',
```
alejoe91: again here tp_rate and accuracy are the same in my opinion

samuelgarcia:
I meant "redundant_units"

spiketoolkit/comparison/groundtruthcomparison.py
```diff
@@ -0,0 +1,392 @@
...
+        GroundTruthComparison.count
+      * compute the confusion matrix .get_confusion_matrix()
+      * compute some performance metric with several strategy based on 
+        the count score by spiketrain
+      * count how much well detected units with some threshold selection
+      * count false positve detected units
+      * count units detected twice (or more)
+      * summary all this
+    """
+    def __init__(self, gt_sorting, other_sorting, gt_name=None, other_name=None,
```
alejoe91: `other_sorting` maybe could be `output_sorting` or `test_sorting`?

Not sure :)

alejoe91:
I like `redundant_units`

spiketoolkit/comparison/groundtruthcomparison.py
samuelgarcia: Yes you are right.
Do we have the same definition as Jeremy website ?

spiketoolkit/comparison/groundtruthcomparison.py
samuelgarcia: Also in our case we don't have mertics taht is penalized by cl ?

samuelgarcia:
OK I make the change obersplitted > redundant

About other_sorting > test_sorting I would prefer tested_sorting. But other is OK for me.
What do you think ?
I don't like so much "output_sorting". Sorry.

alejoe91:
ss_sorting --> spike-sorted_sorting?
I like test_sorting too. tested_sorting so and so

samuelgarcia:
Lest go for tested_sorting

alejoe91:
Also for performance I think we should keep (https://github.com/SpikeInterface/spiketoolkit/issues/123):

- accuracy
- specificity
- precision
- false discovery rate
- miss rate
- misclassification rate

False positive rate according to wiki is:
fp / (fp + tn)
So we can't compute it and we should drop it.

Tp_rate coincide to sensitivity (and we compute it wrongly right now).

fn_rate == miss_rate

Ok? Let's finish up this semantics in this PR so comparisons are fixed once and for all :)

samuelgarcia:
Note : we also need to make a big clean in SortingComaprison : not done at all.
There are still mistake there.

samuelgarcia:
spikeforest have also recall.

alejoe91:
@samuelgarcia to have exactly the same as spike-forest we just have to rename:

sensitivity --> recall (they are the same thing)

samuelgarcia:
Why do you want to remove fp_rate define fp / num_gt ?

samuelgarcia:
@alejoe91 : I have modified semantic and formula in get_performence ?
Could you carrefully have a look.
Formula are now easier to read.

I you validate this I will propagate changes to other function (get_well_detected_units , ...)

alejoe91:
because it's wrong according to wikipedia https://en.wikipedia.org/wiki/Sensitivity_and_specificity :)

alejoe91:
"fall-out or false-positive-rate: FP / (FP + TN)"

As you suggested I think we have to be cristal clear here.

alejoe91:
plus, precision exactly tells you how many TP / FP are found

alejoe91:
@samuelgarcia new semantics for performance looks good :)

samuelgarcia:
@alejoe91 I think this is finish.

But...
..this PR was triggered by @mhhennig because there were strange formula in :
SortingComparison.get_false_positive_fraction()
SortingComparison.get_false_negative_fraction()

The main problem with theses two functions : they are not bijective,  in sens if you invert sorting1 and sorting2 you will not have the same result.

@magland : do you use this internally in spikeforest :
So what can we do here : remove then or find a better denominator ?
In a non GT comparison I don't anderstand the meaning of false_negative and false_positve.
I would vote for remove.

alejoe91:
@samuelgarcia

I think we can keep get false positives and false negatives fraction in the ground truth comparison.

I agree it doesn't make sense to have them in the SortinfComparison.

For the gt case, we can normalize with n_gt. Also nomenclature is fine because the 'official' names are 'rate', not 'fraction'. We can extend the docs and explain what they return

samuelgarcia:
Are you suggsting me to move : 
get_false_positive_fraction() and get_false_negative_fraction() to GroundTruthComparison ?

I think it is a bad ID because GroundTruthComparison.get_performance(self, method='by_spiketrain') already give this. And many more. Why should we have two function for two columns ?

I think it overload GroundTruthComparison without real usecase.
The actual state of GroundTruthComparison is quite clear.
No ?

I propose to leave this two function in SortingComparison but commented. For the moment.
So we could merge this PR soon.
If one day we really need this function we can implemented this again.

samuelgarcia:
@alejoe91 : could you help me here a have a look into comparison_example.ipynb to see what we put/remove in this example ?

alejoe91:
Ok what do you suggest?

samuelgarcia:
Yes the structure is OK.

Do not spend to much time on itbecause I think we should move to sphinx galery soon to avoid NB in git.
And in case of sphinx galery Ithink you should go several sub example files.

samuelgarcia:
@alejoe91 I propose to merge this now.
You will be able to make the example directly in master with another PR.
Ok ?

alejoe91:
ok merged

samuelgarcia:
# 🍾

Issue #68, opened 2019-03-13
Flat dictionaries for HS and tridesclous


martinosorb:
As mentioned in the call, we should "expose" the parameters to the user. An option that I would suggest is to use a hybrid approach. Parameters that we consider important (i.e. the ones that the user almost always will want to change) are in `self.params`. The other parameters are in one or more dictionaries which are themselves in `self.params`:

```python
self._default_params = {
    'important_par1': 0.0,
    'important_par1': True,
    'other_pars': { 'par1': 0.0, 'par2': 0.0 }
}
```
In this way, the user can use `set_params` directly for the main parameters, and, if advanced parameters need to be changed, can go through the more complex procedure of building a new dictionary and passing it to `set_params`.

Just an idea, what do you think?

alejoe91:
@martinosorb 

I think this could be a good idea. 'other_pars' would also be a flat dictionary right?

martinosorb:
Potentially could be nested, that would make it simpler to then pass those arguments. But I can make it flat if you prefer.

@mhhennig @colehurwitz31 can we agree on what parameters should be considered "important" and "often tuned by the user" in HS? Definitely `alpha, bandwidth, n_jobs` on the clustering side (and I suggest setting a default bin_seeding=True perhaps?) and `threshold` for the detection. Cutout start, end should be made fixed once HS is turned into using milliseconds as units.
What about probe params?

mhhennig:
For detection, I agree `threshold` matters most. There is also:

`maa`: minimum spike area (cumulative), to exclude 1 frame baseline deviations
`maxsl`: maximum width of negative deflection
`minsl`: interval to compute `maa` over
`ahpthr`: value the repolarisation has to exceed to be counted as spike
`inner_radius`: spatial radius used to compute centre of mass
`neighbor_radius`: maximum spatial spread of signals, for excluding duplicates
`cutout_start`: frames before peak to include in cutouts
`cutout_end`: frames after peak to include in cutouts

Spatial parameters should be in um (need to adjust the biocam probe here), temporal parameters best in ms, to be converted to frames using the sampling rate internally.

Clustering:
`pca_ncomponents`: default 2
`alpha`, `bandwidth`: perhaps default to about 8, 8um
`cluster_subset`: important for performance, can be set to 200,000 or so, so that only a subset of events will be used to form clusters

alejoe91:
@samuelgarcia any update on this?

samuelgarcia:
[...]

PR #122, opened 2019-05-16
flat dictionary for hs

martinosorb:
Cf. #68. I went for fully flat here, which is perhaps a bit confusing to the user who doesn't know what needs to be tweaked, but I suppose that's more of a problem of HS rather than spiketoolkit.


samuelgarcia:
This look good.
Do you plan to include filtering capability in HS2 in this PR ?
Or is it ready to merge ?

martinosorb:
I don't know about the filtering, I'll need to ask Cole and Matthias.

alejoe91:
You can check how it's implemented in mountainsort4. I think it's a great addition and the default could be False

alejoe91:
@martinosorb I'll merge this and we can make a new PR for the filter

Issue #124, opened 2019-05-22
Question/feedback how SNR is compute


samuelgarcia:
@alejoe91 @colehurwitz31 @mhhennig 
I reading the code on how the the SNR is computed in qualitymetrics.

Some questions/feedback:
  * I don't see any filtering step, how can estimate the SNR on the full band ?
  * the std is not robust at all to estimate the noise, we need to use MAD. The std is bias with the 
    spike rate.
  * the full signal duration is used to estimaded the noise (std/mad), it is useless if the noise is 
    stable, 30s of signal is enough because mad is robust. It will be faster.
  * the amplitude is computed as max(template) - min(template), I think we should keep 
    max(template) is peak is positive and abs(min(template)) is peak is negative
  * for amplitude the recording need also to be taken on high pass signal other wise it is a 
    mistake.
  * template are computed with all spikes, this is tottaly unsane for memoery and CPU.
  * template are computed with np.mean they must be computed with np.median.

I think we need to fix this before the release. The SNR estimation is weak in actual state.

colehurwitz31:
I agree on these issues. I did not implement the templates or the SNR so I am not sure completely how they work. Alessio would have a better idea of why he implemented them the way he did.

alejoe91:
Hi Sam :) @magland implemented it. I agree with those changes.

- We can add an optional filtering step, which can be only HP
- I agree MAD should be used 
- we can use 30 second snippet
- Ok for using: abs(min(template))
- Templates can be easily computed with less waveforms by using the arg `max_num_waveforms` (already implemented)
- yeah median should be more robust

Let's implement those changes ;)

samuelgarcia:
I am not sure to have time just now to do it.
I would prefer fisrt finish the actual PR.
Clean launcher and make more easier to use.







And get better benchmark for tdc.... :(

colehurwitz31:
@alejoe91 Could you possibly implement these changes?

On a separate note, the GUI is coming along and an alpha version should be available soon for people to try.

alejoe91:
@samuelgarcia BTW right now actually the snr was estimated only on 1000 samples!

I will set 10 seconds as default

samuelgarcia:
Oups. I di not see it.
Where is it in the code ?

alejoe91:
https://github.com/SpikeInterface/spiketoolkit/pull/126

alejoe91:
[...]

PR #126, opened 2019-05-23
Fixes to SNR computation

alejoe91:
- default uses MAD, but STD can be used as well
- template can be computed as mean or median
- SNR is computed on the electrode with largest peak
- peak is computed as max(abs(min(template)))


alejoe91:
https://github.com/SpikeInterface/spiketoolkit/issues/124

spiketoolkit/validation/qualitymetrics.py
```diff
@@ -2,7 +2,8 @@
 import spiketoolkit as st
 
 
-def compute_unit_SNR(recording, sorting, unit_ids=None, save_as_property=True):
+def compute_unit_SNR(recording, sorting, unit_ids=None, save_as_property=True, mode='mad',
+                     seconds=10, max_num_waveforms=1000, filter=False, freq_min=300, freq_max=6000):
```
samuelgarcia: with median 300 is enough.

spiketoolkit/validation/qualitymetrics.py
```diff
@@ -25,28 +38,85 @@ def compute_unit_SNR(recording, sorting, unit_ids=None, save_as_property=True):
...
+        Noise levels for the different channels
+    max_channel_idx: int
+        Index of channel with largest templaye
+
+    Returns
+    -------
+    snr: float
+        Signal-to-noise ratio for the template
+    '''
+    snr = np.max(np.abs(np.min(template[max_channel_idx]))) / channel_noise_levels[max_channel_idx]
```
samuelgarcia: In case of positive spike I would do
max( np.max(np.abs(np.min(template[max_channel_idx]))), np.max(template[max_channel_idx]) ) / ...

spiketoolkit/validation/qualitymetrics.py
samuelgarcia: To be sure that function is correctly used I would put filter=True.

So if you known what you do you can always disable but if you don't known it is better to be safe.

samuelgarcia: No sorry i misunderstood.
I would put **already_filtered=False** or at least **fileterd=False**
filter is ambiguous because it can both mean already filtered or apply filter

samuelgarcia:
Thanks.
This look good.
I made some comments.
tell me what you think.

spiketoolkit/validation/qualitymetrics.py
alejoe91: I think 300 or 1000 doesn't change much in terms of time. It's fast anyways :)

spiketoolkit/validation/qualitymetrics.py
alejoe91: I need to change the get_unit_max_channel as well, because it uses the same.

maybe just a np.max(np.abs(template[max_channel_idx]))??

spiketoolkit/validation/qualitymetrics.py
alejoe91: Mmm, I think it's better to filter outside. Extracting waveforms/templates on a FilterRecordingExtractor is very bad, because it's lazy and it filters all snippets.

I agree, but first we should add a `cache` arg to the bandpass_filter. If operations like get_waveforms etc. are used the `cache=True` should be used so a big filter is done once only?

Agree? I can make it today

alejoe91: done.

I think `filter` is clear enough with the docs, personally. I would leave it false by default (by the time you run spike sorting you probably have filtered your data already...).
Or we can call it apply_filter :)

alejoe91: used `apply_filter`

samuelgarcia:
Totally perfect.

(Except the apply_filter=False by default but I can get used to it....)

PR #127, opened 2019-05-24
Enhance environ path for KS, KS2 and IronClust

samuelgarcia:
Enhance environ path for KS, KS2 and IronClust

### DIFF
--- a/spiketoolkit/sorters/ironclust/ironclust.py
+++ b/spiketoolkit/sorters/ironclust/ironclust.py
@@ -68,7 +68,14 @@ class IronclustSorter(BaseSorter):
 
     @staticmethod
     def set_ironclust_path(ironclust_path):
+        os.environ["IRONCLUST_PATH"] = ironclust_path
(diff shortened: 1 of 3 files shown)


samuelgarcia:
This PR help for environement path for KS, KS, IronClust:
  * modify the class attr **installed**
  * check in params when the path is present
  * put the path with os.environement

alejoe91:
ok great

alejoe91:
so if I pass the path to `run_kilosort2` it saves it as os variable? does it work on Windows as well?

samuelgarcia:
Not tested on windows. Nor really tested on linux but the logic was quite not good because installed was not updated and params not checked on this.

samuelgarcia:
I think we can merge.
Thing like that
```os.environ["IRONCLUST_PATH"] = ironclust_path```
 will not be persistent but at least during the python run we have a consistent path between class attr and the local env.

alejoe91:
ok

PR #129, opened 2019-05-28
Force tridesclous=1.1.0

samuelgarcia:
I have released tridesclous to 1.1.0.
So I force travis.
We should also do the same for HS2 and mountainsort.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   - pip install Cython
   - pip install h5py # for klusta
   - pip install click klusta klustakwik2
-  - pip install https://github.com/tridesclous/tridesclous/archive/master.zip
+  - pip install tridesclous==1.1.0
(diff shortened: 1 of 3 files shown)


samuelgarcia:
Travis fail because of moutainsort because of ml_ms4alg pypi version.
See https://github.com/magland/ml_ms4alg/issues/17

I think we can merge this.

samuelgarcia:
I think we should reverse the order of import in __ini__.py
```
from . import postprocessing
from . import sorters
from . import preprocessing
from . import comparison
from . import validation
```
should be
```
from . import postprocessing
from . import preprocessing
from . import sorters
from . import comparison
from . import validation
```

no ? I think this cause travis bug.

alejoe91:
yeeeeeeeee :) merging!

PR #132, opened 2019-05-29
Filters for HS2

martinosorb:
This adds a bandpass filter. Note: I did not test this, only copied from MountainSort.

Additionally, it adds the Rescale Traces filter. This measures certain quantiles of the activity and rescales the traces in order to make the detection independent of the units used for voltage.


colehurwitz:
As Matthias was suggesting, do you think you could change the name to something like: normalize_by_quantile or something?

Also, do you think the new preprocessing tool can return an object of: TransformTracesRecording? You can calculate the scalar and offset and then pass those into the TransformTracesRecording object that is to be returned without making a new class.

martinosorb:
Matthias and I settled for just normalize. Let me know if there are objections.

alejoe91:
To me normalize mean (signal - mean(signal))/std(signal)..

I agree that maybe normalize_by_quantile is more specific here.

@martinosorb @colehurwitz what do you guys think?

martinosorb:
I thought NormalizeByQuantileRecording sounded awkward.

colehurwitz:
I think it is awkward, but probably the best choice for now. We can change the name later if we come up with a better one.

alejoe91:
@martinosorb I agree but that what it does :)

martinosorb:
Ok changed it

colehurwitz:
Ready to merge?

martinosorb:
As far as I'm concerned yes

mhhennig:
The name is horrible, but ok...

colehurwitz:
Haha, do you have a suggestion Matthias? I think it is not great as well

mhhennig:
No better idea right now, let's merge (but we should not change it later so we don't break dependencies).

mhhennig:
All right, the class could be called: `NormalizedExtractor`
Then we can add methods, at the moment it is just `normalize_by_quantile`

alejoe91:
I think there could be a place for another `NormalizeExtractor` class that does the z-transform. Also the class name is not available to the user, who will always call the `normalize_by_quantile` function :)

I'll merge it!

samuelgarcia:
NormalizeExtractor  : (sig-mean(sig))/std(sig)

RobustNormalizeExtractor  : (sig-median(sig))/mad(sig)

PR #133, opened 2019-05-29
Still improve the logic of global path in KS, KS2 and Ironclust.

samuelgarcia:
Still improve the logic of global path in KS, KS2 and Ironclust.
This is because path can be change at everytime in two way (by call function or os.environ)


samuelgarcia:
Sorry to change this again.
The logic was still wrong.

samuelgarcia:
Merge this.

PR #135, opened 2019-05-30
Fixes to export to phy pc_feature_ind

alejoe91:
- Fixes a bug in setting the right pc feature indexes for the template gui
- Clean up comments


alejoe91:
@samuelgarcia I don't know why klustakwik2 is failing. 
I tested the same .travis install on a new environment and it works...

PR #137, opened 2019-06-02
Completed read the docs with updated comparison and validation

alejoe91:


samuelgarcia:
@alejoe91 : thanks you for this.
This is excelent.

If I have time I will write more detail on sorter submodule, we should emphasize more the fact that they are wrapper around external codes (bash, matlab or python diirectly).

PR #139, opened 2019-06-04
MultiSortingComparison bug fix

alejoe91:
When looking for the maximum edge (agreement) between units, the order should be of the original graph needs to be kept, otherwise the edges might not be found in the original graph

### DIFF
--- a/spiketoolkit/comparison/multisortingcomparison.py
+++ b/spiketoolkit/comparison/multisortingcomparison.py
@@ -148,7 +148,10 @@ class MultiSortingComparison():
                     for n2 in nodes:
                         if n1 != n2:
                             if (n1, n2) not in edges and (n2, n1) not in edges:
-                                edges.append((n1, n2))
+                                if (n1, n2) in self.graph.edges:
+                                    edges.append((n1, n2))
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Did not anderstand the why.
I trust you I did not use and inspect MultiSorting yet.

alejoe91:
Basically first we build the graph in which each node is the sorter name + unit (e.g. TDC-4). The edges connect units and their weight is the agreement score (e.g. KS-4_TDC-1). 
Then spiketrains are built as follow: for each unit, all the edges are gathered and the final spike train is made up with only the true positives spikes between the comparison with the largest agreement. 
When looping through the nodes, however, the order of building the edges might be reversed, so i check which edge is in the original graph...

It's a bit complicated but I found this bug when working on my tetrodes data. The consensus-based cleans up a lot of noisy units! :)

samuelgarcia:
OK thank.
I am sure to match totally on the idea of concensus method.
In my mind if 3 sorter do not resolve collision and 1 do resolve it then the concensus is to remove hard to find spike, no ?
Lets talk about this in the meeting.

PR #140, opened 2019-06-04
Added correct GUI features

colehurwitz:
Added a bunch of gui features to different sorters and to the post-processors. Right now, I listed Threshold_Min_SNR as not installed because I haven't been able to put that into the GUI very well yet. Will work on that.

Let me know what you guys think, but these were mostly just changes for the GUI

### DIFF
--- a/spiketoolkit/sorters/spyking_circus/spyking_circus.py
+++ b/spiketoolkit/sorters/spyking_circus/spyking_circus.py
@@ -21,7 +21,6 @@ class SpykingcircusSorter(BaseSorter):
 
     sorter_name = 'spykingcircus'
     installed = HAVE_SC
-    SortingExtractor_Class = se.SpykingCircusSortingExtractor
 
     _default_params = {
         'probe_file': None,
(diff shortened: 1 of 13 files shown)


spiketoolkit/sorters/spyking_circus/spyking_circus.py
```diff
@@ -57,7 +58,7 @@ def _setup_recording(self, recording, output_folder):
         if p['probe_file'] is None:
             p['probe_file'] = output_folder / 'probe.prb'
             se.save_probe_file(recording, p['probe_file'], format='spyking_circus',
-                               radius=p['adjacency_radius'], dimensions=p['electrode_dimensions'])
+                                radius=p['adjacency_radius'], dimensions=p['electrode_dimensions'])
```
alejoe91: can you fix this indentation? `radius` should start as `recording`

samuelgarcia:
Why did you put back SortingExtractor_Class on classes ?
I put at the very beginning of classes but it was not flexible enough.
I removed this from each sorter and replace by aclass method wich is more convinient:
```
    @staticmethod
    def get_result_from_folder(output_folder):
```

samuelgarcia:
See IronclustSorter.

colehurwitz:
Ah, I put that back because I thought it would be useful for the GUI to know what format the sorter wrote to. I didn’t end up using it though so I can remove that

PR #142, opened 2019-06-06
Adding detailed gui params to sorters [WIP]

colehurwitz:
I will be adding gui params in this pull request so don't merge yet please. Starting with Klusta and working my way up the list for the ones I haven't done yet.

Let me know what you think about my gui params for each one if you get the chance.


samuelgarcia:
You can add a [WIP] prefix for the title to warn about the no merging yet.

colehurwitz:
How the heck does one README change pass and another fail...

PR #152, opened 2019-06-11
Changed default value of HS2 to match spikeforest

colehurwitz:


samuelgarcia:
Also figting to get better results ?

colehurwitz:
Haha, yes. We are are getting there!

PR #153, opened 2019-06-11
Change logic to avoid to force .dat for binary

samuelgarcia:
In actual implementation a binary file for klusta MUST be ".dat" now it is can be any extention (.raw, .bin)

### DIFF
--- a/spiketoolkit/sorters/klusta/config_default.prm
+++ b/spiketoolkit/sorters/klusta/config_default.prm
@@ -3,7 +3,7 @@ experiment_name = r"{}"
 prb_file = r"{}"
 
 traces = dict(
-	raw_data_files=[experiment_name],
+	raw_data_files=[r"{}"],
 	voltage_gain=1.,
 	sample_rate={},
 	n_channels={},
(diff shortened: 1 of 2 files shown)


colehurwitz:
That is a nice change!

samuelgarcia:
I thought it was simple but travis is unhappy...

samuelgarcia:
I fix it.
But finally klusta really need a file with .dat extension.
This restriction is really annoying.
I propose to merge and I will be back on this soon.

alejoe91:
should we rename this https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extraction_tools.py#L224 
`write_binary_file` (dropping the dat) accordingly?

samuelgarcia:
Not not merge, there is sttill a bug.

samuelgarcia:
Yes I think we could rename this function.
dat is not official at all many format use raw or bin.

klusta working only with dat is really a pain.

samuelgarcia:
Finally there is no bug. i merge this.

PR #154, opened 2019-06-12
 Change logic to avoid to force .dat for binary

samuelgarcia:


samuelgarcia:
idelaly I would like to merge this before the release.

colehurwitz:
Sounds good. Btw, Alessio and I are adding sampling frequencies to current SortingExtractors. What do you recommend for formats that don't have that information?

I was thinking an optional sampling_frequency argument can be added to those sorting extractors like:
MDASortingExtractor(firings, sampling_frequency=None)

samuelgarcia:
I think it is better to have explicitly self._sampling_frequency = None for theses format.

The only consequence is that the user will have to provide manually for SpikeComparison otherwise it will raire an error.

samuelgarcia:
This is a very small change I merge it.

PR #155, opened 2019-06-12
postprocessing fixes

alejoe91:
some cleaning and fixes to postprocessing module


colehurwitz:
What 'property' values does extract_property_from_maximum_channel support?

Can you list the available properties in the docstring?

alejoe91:
it's any channel property :) in general it would be group

colehurwitz:
Right, but channel properties are finite so not any string they put in will work. Off the top of my head, we have 'group', 'location'... etc.

colehurwitz:
Wait, I think I am misunderstanding the function

colehurwitz:
What does this function do?

alejoe91:
they are not finite. One can load any channel property.

So what it does is to extract the property from the channel with the largest peak and save the same property as sorting property

colehurwitz:
Sorry, so what channel properties would you want to store as unit properties?

alejoe91:
'group', 'brain_area', whatever really :)

alejoe91:
e.g. say you have group information, but you don't spike sort by group. Then this function can be used to assign each unit to the right group ;)

colehurwitz:
Could we potentially change the name? I think extract_property_from_maximum_channel doesn't really give an idea of what it does.

Maybe something like: 
```
set_units_properties_by_max_channels_properties
```

 or something?

colehurwitz:
Also, I think  get_unit_template --> get_units_templates since it is multiple units.

colehurwitz:
And  get_unit_max_channel --> get_units_max_channels

colehurwitz:
Also, maybe compute_pca_scores --> comput_unit_pca_scores

alejoe91:
You can go ahead and make the changes :) all this functions are also used by other functions + notebooks + docs

colehurwitz:
Ok, I will change all of those

colehurwitz:
You can merge this first

colehurwitz:
I just think the current names don't match the intended use and will confuse people who are unfamiliar with the functions (e.g. me).

alejoe91:
I think though it should be unit_templates etc. Not units

alejoe91:
Similar to get_channel_groups, not channels_groups

colehurwitz:
I think it has to be singular or plural based on what it does. So get_units_templates is more accurate to what the function does.

alejoe91:
It can return a single template as it takes the unit_ids arg. I would prefer get_unit_...s notation.

PR #156, opened 2019-06-12
Name changes to postprocessing modules

colehurwitz:
I think I propagated everything correctly (with find all and replace all). Maybe someone could check to make sure.

We can discuss name changes here so I won't merge yet.


colehurwitz:
I can change it to singular if you want Alessio... You may be right

colehurwitz:
I changed to singular notation. I think you are correct. That means that only some of the function really had their name changed..

PR #159, opened 2019-06-13
Strong bug in get_redundant_units

samuelgarcia:
I have just discover a string bug in get_redundant_units
The logic was totally false.
Sorry this is my fault.

This should be fixed now.
We will need to release again before the meeting for this.

@alejoe91 
@colehurwitz

### DIFF
--- a/spiketoolkit/comparison/groundtruthcomparison.py
+++ b/spiketoolkit/comparison/groundtruthcomparison.py
@@ -1,4 +1,5 @@
 from .basecomparison import BaseComparison
+from .comparisontools import compute_agreement_score
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


alejoe91:
@samuelgarcia no problem. I would recommend waiting for next week for the next release. We might do other fixes in the meanwhile :)

Ready to merge this?

colehurwitz:
Hmm, maybe worth releasing now if it is a strong bug. Could be a bit of a hotfix.

samuelgarcia:
You can merge if you want but please read carrfully the proposed solution.

alejoe91:
I think we might need a `min_redundant_agreement`. In fact the `min_agreement` could be quite high, but redundant units probably match a subset of the spikes. What do you think?

samuelgarcia:
I think you are right but it make the API more complexe with more kargs.
And so print_summary and collect_result will be another additional args

alejoe91:
Yes but otherwise I think it would not spot redundancy...

samuelgarcia:
@alejoe91 : I have included it.

PR #161, opened 2019-06-17
For tridesclous release 1.2.1

samuelgarcia:


samuelgarcia:
@alejoe91 @colehurwitz after this I would be happy to have a new tag :)

PR #162, opened 2019-06-18
Big agregate of all individual results.

samuelgarcia:
New idea.


colehurwitz:
Nice idea! I like that you can get all at once.

samuelgarcia:
Can I merge this ?

colehurwitz:
Yes, I think it is good

PR #163, opened 2019-06-18
Same changes as in SpikeForest

yger:
I copied the changes submitted to SpikeForest, in order for the two repos to be in sync. Note that current version of SC is now 0.8.0, available with pip/conda/docker. These changes are not meant to turn the code into a pure automatic software. They are just slightly decreasing the number of clusters, and further investigation should be performed to turn it into a pure automatic workflow.

### DIFF
--- a/spiketoolkit/sorters/spyking_circus/config_default.params
+++ b/spiketoolkit/sorters/spyking_circus/config_default.params
@@ -38,8 +38,7 @@ make_plots     =            # Generate sanity plots of the averaged artefacts [N
 
 [whitening]
 chunk_size     = 30         # Size of the data chunks [in s]
-safety_time    = auto          # Temporal zone around which templates are isolated [in ms, or auto]
-temporal       = False      # Perform temporal whitening
+safety_time    = auto       # Temporal zone around which templates are isolated [in ms, or auto]
 spatial        = True       # Perform spatial whitening
 max_elts       = {}         # Max number of events per electrode (should be compatible with nb_elts)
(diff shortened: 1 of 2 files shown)


colehurwitz:
Great! Thanks for propagating the changes over to spiketoolkit. I think this can be merged as it passes all tests, right @samuelgarcia @alejoe91 ?

alejoe91:
@yger Thank you for the changes! Merging :)

samuelgarcia:
@yger : what you can also do to avoid to change code and params in 2 places is to change the spikeforest wrapper on top on the spiketoolkit wrapper.

I did this for tridesclous (and @colehurwitz did the same for heardingspike2), so now I only need to change params in spiketoolkit + change version in spikeforest which is more simple.
Have a look of tridesclous wrapper in SF.

Pierre : we did some comparisons which we will present t next Tuesday in Edimburgh.
Unfortunatly it was done with previous version of SC. So results should be not as good as with this version.
Nervertheless, the presentation is not about comparing sorter but to show how easy you can do it with spiketoolkit.

yger:
Ok, I'll have a look. @samuelgarcia Comparison on what? Performances, speed, as in SpikeForest? If performances, with greedy merges or not? As said, I really think a global framework to compare sorters is a great idea, and I'll try to be more involved, but putting numbers is hard, since every sorters have somehow hidden assumptions.

samuelgarcia:
We try to put several numbers.
That is the topic.

No greedy merge, we do comparison on raw output.

PR #168, opened 2019-06-20
New default parameters for next release

mhhennig:
Now performs scaling and bandpass by default as this is required in almost all cases. This can be turned off for Biocam data for a slight speed increase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
   - pip install https://github.com/SpikeInterface/spikeextractors/archive/master.zip
   - pip install .
   - pip install pytest==4.3
-  - pip install numpy scipy h5py six cython tqdm click # for klusta
+  - pip install numpy==1.16.2 scipy h5py six cython tqdm click # for klusta
   - pip install klusta
   - pip install klustakwik2==0.2.6
   - pip install tridesclous==1.2.1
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Have you seen error on travais about numpy.quantile ?

mhhennig:
I'm trying to fix, it is the numpy version. I don;t understand though why Travis fails now.

alejoe91:
Sometimes hs fails.. don't know why. You can trigger the build it again from travis

mhhennig:
I just replicated the test locally, with the same packages, and cannot find the reason it fails. Help please :(
One thought, is the test data generated randomly, with a different seed each time? Is it possible it fails as it cannot detect any units? Can anyone point me to where this is done?

colehurwitz:
I can try to help! I’ll go onto the branch and try soon

samuelgarcia:
One travis is passing and the other is wrong.

I think we can merge this. it is only parameters.

mhhennig:
Oh, passing. Can we fix a random seed for the test data, I think this will solve this?

colehurwitz:
I tried to fix the random seeds, but maybe I missed some! Would be good to fix all of them

samuelgarcia:
Do we more random else where than toy example ?

colehurwitz:
Let's merge? I can't find any other randomization

alejoe91:
yes it passes :)

PR #170, opened 2019-06-21
Proper names on confusion matrix and option to turn off count text.

samuelgarcia:


colehurwitz:
I noticed that the confusion matrices were quite hard to read. Will this fix that?

samuelgarcia:
It remove text optionally usefull for big matrices.
I am unlucky with travis.
I hung on hs. My guess is that there is compilation imcompatibility somewhere.
Is HS compiled on travis ?

alejoe91:
it looks fine on travis
I suggest we merge :)

samuelgarcia:
It is because I relaunch it.
We now have some random fail (hung) at hs steop. My guess is compilation problem.

PR #174, opened 2019-06-27
Use spikesorters package for sorters

alejoe91:
Remove all sorters and import `spikesorters` module in sorters submodule.

@samuelgarcia @magland @colehurwitz  should we add a hard dependency on spikesorters at this point?


samuelgarcia:
@alejoe91 : what do you call hard depency ?
Is it with a ficed version or not ?

I think yes we must have in setup a dependency on 'spikesorters' but without version.

You fogot to add the dependency in travis. I think.

alejoe91:
@samuelgarcia 

The tests here will be only for preprocessing (TODO), comparison, study, postprocessing (TODO), and validation (TODO).

Removing all the spike sorting specific installations from the travis

alejoe91:
So spikesorters is up and running. This is up and running too. Can I merge? @samuelgarcia @colehurwitz

colehurwitz:
If it works, I am good to merge!

Issue #128, opened 2019-05-24
Sorting comparison: Allow milliseconds time jitter instead of the number of samples (delta_tp) 


jamesjun:
Sorting comparison class should allow milliseconds input instead of allowing the number of samples by passing delta_tp. Currently, the sorting comparison doesn't take different sampling rate into account, penalizing higher sampling rate recordings.

alejoe91:
@jamesjun that's a good point.
In order to do that we have to add a `self.sampling_frequency()` variable to all extractors. Right noe SortingExtractor don't have sample rate information, but I think it's a great addition.

@colehurwitz @samuelgarcia what do you guys think?

colehurwitz:
I don't think we should add sampling_frequency to the sorting extractor. We already kinda talked about this and we agreed that it should be a part of the recording extractor only (which it is) since the sorting extractor doesn't necessarily know about the recording.

I would recommend adding a sampling frequency argument to the function instead.

alejoe91:
I think the spike sorter knows the sampling frequency. It's a required parameter to all sorters and if we don't have it in the sorting extractors we will always have to tag along the recording extractor.
I think it doesn't harm to add it as a property

colehurwitz:
Hmm.. I am not sure the sorting extractor should have it as a property. Right now we only ever deal with unit properties, so there really isn't a concept of global properties. 

I think putting sampling frequency as an argument makes more sense because if you are running the sorter (and it returns the sorting extractor) than you definitely have access to the recording extractor already. And for the times that you are using a sorting extractor without using a sorter, you would have to manually set that property before using the comparison which is clunky.

alejoe91:
I don't agree. I think it's muuuuch easier to have a get sampling frequency internally.

Think you want to load a sorting output later and compute something: you don't want to track back and load the original recording...

colehurwitz:
If sampling frequncies are not stored in the sorting files then they shouldn’t be stored in the sorting extractors. Do any formats have access to this information? 

If no formats have this information, then I am not sure what adding this to the sorting extractor gains you because you need to know the sampling frequency from the recording anyways.

colehurwitz:
It’s not like we ever save a sorting extractor object so the information about sampling frequency would have to be known by the user each time and added

samuelgarcia:
@jamesjun  : I agree that this params should be in ms.
Indeed a default params is never changed because we all forgot it so having it in ms is safer.

[...]

PR #177, opened 2019-06-28
Matching based on time instead of number of frames

alejoe91:
This fixes this issue https://github.com/SpikeInterface/spiketoolkit/issues/128

@jamesjun @magland when we merge we should propagate this change in SpikeForest

### DIFF
--- a/spiketoolkit/comparison/basecomparison.py
+++ b/spiketoolkit/comparison/basecomparison.py
@@ -1,140 +1,79 @@
-import spikeextractors as se
 import numpy as np
-
-from .comparisontools import (count_matching_events, compute_agreement_score,
-                                                do_matching, do_score_labels,  do_confusion_matrix)
+from .comparisontools import (do_matching, do_score_labels, do_confusion_matrix)
 
 
 class BaseComparison:
     """
     Base class shared by SortingComparison and GroundTruthComparison
-    
     """
-    def __init__(self, sorting1, sorting2, sorting1_name=None, sorting2_name=None, delta_frames=10, min_accuracy=0.5,
-                n_jobs=1, verbose=False):
-        self._sorting1 = sorting1
-        self._sorting2 = sorting2
+    def __init__(self, sorting1, sorting2, sorting1_name=None, sorting2_name=None, delta_time=0.3, min_accuracy=0.5,
+                 n_jobs=1, verbose=False, sampling_frequency=None):
+        self.sorting1 = sorting1
+        self.sorting2 = sorting2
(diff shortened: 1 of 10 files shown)


spiketoolkit/comparison/basecomparison.py
```diff
@@ -1,140 +1,79 @@
...
-    
     """
-    def __init__(self, sorting1, sorting2, sorting1_name=None, sorting2_name=None, delta_frames=10, min_accuracy=0.5,
-                n_jobs=1, verbose=False):
-        self._sorting1 = sorting1
-        self._sorting2 = sorting2
+    def __init__(self, sorting1, sorting2, sorting1_name=None, sorting2_name=None, delta_time=0.3, min_accuracy=0.5,
+                 n_jobs=1, verbose=False, sampling_frequency=None):
+        self.sorting1 = sorting1
+        self.sorting2 = sorting2
```
samuelgarcia: Yeah!

spiketoolkit/comparison/basecomparison.py
```diff
@@ -1,140 +1,79 @@
...
     def get_labels2(self, unit_id):
-        if unit_id in self._sorting1.get_unit_ids():
+        if unit_id in self.sorting1.get_unit_ids():
             return self._labels_st1[unit_id]
         else:
             raise Exception("Unit_id is not a valid unit")
 
-
-
-    def plot_confusion_matrix(self, xlabel=None, ylabel=None, ax=None, count_text=True):
```
samuelgarcia: No! Please keep it a bit more.
Please. i need a bit of time to see spike widget.
Please don't make unrelated to PR change.

samuelgarcia:
@alejoe91 : you make too much change in the same PR.

Everything about the delta_time is OK for me.

but you should target only one thing at a time :)

I really want to have comp.plot_confusion_matrix()!!
Even if the code if elsewhere having it directly in method is cool.

What will be the dependency between spiketoolkit and spikewidget ?
I really plan to have plot facilitis in spiketoolkit (in comparison and in the study module) even if it call another module (widget).

spiketoolkit/comparison/basecomparison.py
alejoe91: added back :)

alejoe91:
@samuelgarcia I agree. I cleaned the PR and only targeter delta_frames. Also, as suggested, I rmeoved get_sorting1 and exposed sorting1 as member variable. Is that ok?

alejoe91:
can we merge this? @samuelgarcia @magland

PR #190, opened 2019-07-01
Make miclassification error optional (it can be very time consuming)

alejoe91:
- makes labeling optional (True by default for groundtruth)
- makes misclassification error computation optional (False by default)


samuelgarcia:
OK done before, I try start it :)
@alesio : We also need to change all print_performance depending on this optional.

alejoe91:
Right :) sorry I merged to make a PR to SF!

PR #191, opened 2019-07-01
Some other change about the compute_misclassification option.

samuelgarcia:
* label > compute_label
  * template_txt_performance depend on self._compute_misclassification
  * compute_misclassification=True for GroundTruthSoringComparison.

### DIFF
--- a/spiketoolkit/comparison/basecomparison.py
+++ b/spiketoolkit/comparison/basecomparison.py
@@ -7,7 +7,7 @@ class BaseComparison:
     Base class shared by SortingComparison and GroundTruthComparison
     """
     def __init__(self, sorting1, sorting2, sorting1_name=None, sorting2_name=None, delta_time=0.3, min_accuracy=0.5,
-                 n_jobs=1, verbose=False, sampling_frequency=None, compute_label=True, compute_misclassification=False):
+                 n_jobs=1, verbose=False, sampling_frequency=None, compute_labels=True, compute_misclassification=False):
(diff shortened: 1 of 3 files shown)


samuelgarcia:
@alejoe91 : I really don't like the **label** (now rename **compute_label**). option.
label must be compute anyway for any method in this class.

We remove it some times ago, Ithink we where agrre on this and now it appear again.
What is the use case of compute_label=False ?

compute_misclassification=False I understand totally, this is for SF.

@magland : Important note : I put back **compute_misclassification=True** by default in ST.
I think I prefer ST to be exhaustive by default.
So you have to explicitly compute_misclassification=False in SF code for speedup.

@alejoe91 @magland @colehurwitz @mhhennig  This must be discussed, does default params have to match SF so user can run code and have same results immediatly ?
Or can ST have more option (and longer to compute) for comparison ?

colehurwitz:
I think it is ok to have default params be different in ST than in SF and they can change the defaults to match SF

alejoe91:
Can you call it compute_labels?

samuelgarcia:
Yes of course but first tell me the usecase of the kargs :)

alejoe91:
Compute_labels can be set to false in the sortingcomparison, as only the matching is needed

alejoe91:
Compute_misclassification is set to False in SF :)

samuelgarcia:
OK point for you.
Done already.

samuelgarcia:
Merge when you want.

PR #194, opened 2019-07-03
Use best match instaed of hugarian match for couting in GT context.

samuelgarcia:
@alejoe91 @colehurwitz @mhhennig @magland 

This change the behavior of counting in the ground truth context.
It was discussed in Edimburgh.

Before the counting was done on **_unit_map12** (aka hugarian assignement) now it is done on the **_best_match_units_12** (the best match).

It could change a bit the metrics. 
Are we all agree on that ?

Tell me.


alejoe91:
Yep I think that's what we agreed on :)

colehurwitz:
I want to release new versions today. Should this be included?

alejoe91:
I think so! I would only release extractor toolkit and sorters. Not much updates for the widgets. We'll release next week

@samuelgarcia

samuelgarcia:
If you are happy we can merge when you want.

PR #200, opened 2019-07-08
GUI param change for export to phy

colehurwitz:


alejoe91:
https://github.com/SpikeInterface/spikesorters/pull/39

I think that the same comments hold here. Probably cleaner to implement this in spikely as @samuelgarcia suggested

colehurwitz:
Ok, I will remove these types of params and add them to spikely. Need to make a few pull requests across repos then

PR #204, opened 2019-07-10
Changed common reference to accept only list for reference channels

colehurwitz:
@alejoe91  I didn't change the functionality, but I thought it would be better if reference channels was always a list rather than an integer or a list (better to have one type). If the list is length 1, I just store it as an integer so your old code is compatible. This will also help with future static type checking.


alejoe91:
ok this should work. I will start writing tests in the near future :)

PR #206, opened 2019-07-12
Comparison small fixes

alejoe91:
- fix redundant / false positive units

### DIFF
--- a/spiketoolkit/comparison/groundtruthcomparison.py
+++ b/spiketoolkit/comparison/groundtruthcomparison.py
@@ -28,7 +28,7 @@ class GroundTruthComparison(BaseComparison):
     """
 
     def __init__(self, gt_sorting, tested_sorting, gt_name=None, tested_name=None,
-                 delta_time=0.3, min_accuracy=0.5, exhaustive_gt=False,
+                 delta_time=0.3, min_accuracy=0.5, exhaustive_gt=False, bad_redundant_threshold=None,
                  n_jobs=-1, compute_labels=True, compute_misclassification=True, verbose=False):
         if gt_name is None:
             gt_name = 'ground truth'
@@ -39,6 +39,8 @@ class GroundTruthComparison(BaseComparison):
                                 compute_labels=compute_labels, compute_misclassification=compute_misclassification,
                                 verbose=verbose)
         self.exhaustive_gt = exhaustive_gt
+        if bad_redundant_threshold is None:
+            self._bad_redundant_threshold = bad_redundant_threshold
         self._do_count()
 
     def _do_count(self):
@@ -94,7 +96,7 @@ class GroundTruthComparison(BaseComparison):
 
         if method == 'raw_count':
             perf = self.count
-            
+
         elif method == 'by_unit':
             unit1_ids = self.sorting1.get_unit_ids()
             perf = pd.DataFrame(index=unit1_ids, columns=_perf_keys)
@@ -127,7 +129,7 @@ class GroundTruthComparison(BaseComparison):
             template_txt_performance = _template_txt_performance_with_cl
         else:
             template_txt_performance = _template_txt_performance
-        
+
         if method == 'by_unit':
             perf = self.get_performance(method=method, output='pandas')
             perf = perf * 100
@@ -199,13 +201,13 @@ class GroundTruthComparison(BaseComparison):
         """
         if len(thresholds) == 0:
             thresholds = {'accuracy': 0.95}
-        
+
         _above = ['accuracy', 'recall', 'precision']
-        _below = ['false_discovery_rate',  'miss_rate', 'misclassification_rate']
-        
+        _below = ['false_discovery_rate', 'miss_rate', 'misclassification_rate']
+
         perf = self.get_performance(method='by_unit')
         keep = perf['accuracy'] >= 0  # tale all
-        
+
         for col, thresh in thresholds.items():
             if col in _above:
                 keep = keep & (perf[col] >= thresh)
@@ -223,7 +225,7 @@ class GroundTruthComparison(BaseComparison):
         """
         return len(self.get_well_detected_units(**kargs))
 
-    def get_false_positive_units(self, min_redundant_agreement=0.3):
+    def get_false_positive_units(self, bad_redundant_threshold=None):
         """
         Return units list of "false positive units" from tested_sorting.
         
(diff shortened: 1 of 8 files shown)


samuelgarcia:
@alejoe91 : you are supposed to write your thesis.

Some comments my dear:
  * mixing linter correctiion with important logic is hard is follow in one PR
  * It is better when PR focus on one thing (freq OR fal positive OR linter...)

Otherwise;
  * I don't agree with the change in **get_false_positive_units** in my mind false posistive and the case 
    where **self._best_match_units_21[u2] == -1** with the meaning the unit 2 do not match anything.

spiketoolkit/study/groundtruthstudy.py
```diff
@@ -8,131 +8,139 @@
...
-                                        engine='loop', engine_kargs={}):
+                    engine='loop', engine_kargs={}, debug=False):
         run_study_sorters(self.study_folder, sorter_list, sorter_params=sorter_params,
-                                    engine=engine, engine_kargs=engine_kargs)
+                          engine=engine, engine_kargs=engine_kargs, debug=debug)
 
-    def get_ground_truth(self, rec_name):
-        sorting = se.NpzSortingExtractor(self.study_folder / 'ground_truth' / (rec_name+'.npz'))
+    def get_ground_truth(self, rec_name=None):
+        self.scan_folder()
```
samuelgarcia: I don't think we must scan the folder each time.
This can be slow for very big study.

spiketoolkit/study/groundtruthstudy.py
```diff
@@ -8,131 +8,139 @@
...
+        if len(self.rec_names) > 1 and rec_name is None:
+            raise Exception("Pass 'rec_name' parameter to select which recording to use.")
+        else:
+            rec_name = self.rec_names[0]
+        sorting = se.NpzSortingExtractor(self.study_folder / 'ground_truth' / (rec_name + '.npz'))
         return sorting
-    
+
+    def get_sorting(self, sort_name, rec_name=None):
+        self.scan_folder()
```
samuelgarcia: SAme here

spiketoolkit/comparison/groundtruthcomparison.py
```diff
@@ -222,22 +223,31 @@ def count_well_detected_units(self, **kargs):
...
-        fake_ids = []
+        best_match = list(self._unit_map12.values())
+        false_positive_ids = []
         unit2_ids = self.sorting2.get_unit_ids()
         for u2 in unit2_ids:
-            if self._best_match_units_21[u2] == -1:
-                fake_ids.append(u2)
-        return fake_ids
+            if u2 not in best_match:
+                if u2 not in self.get_redundant_units(min_redundant_agreement):
```
samuelgarcia: This can be slow in a loop.

spiketoolkit/study/groundtruthstudy.py
alejoe91: Ok, but can one chec if the folder has been scanne dor not? I can make a flag :)

Also, why does the static method `setup` returns: `return cls(study_folder)`. I think it could return nothing.

spiketoolkit/study/groundtruthstudy.py
alejoe91: ok

spiketoolkit/comparison/groundtruthcomparison.py
alejoe91: I'll use what you suggested (checking best match 21 and 12)

alejoe91:
> @alejoe91 : you are supposed to write your thesis.
> 
> Some comments my dear:
> 
> * mixing linter correctiion with important logic is hard is follow in one PR
> * It is better when PR focus on one thing (freq OR fal positive OR linter...)
> 
> Otherwise;
> 
> * I don't agree with the change in **get_false_positive_units** in my mind false posistive and the case
>   where **self._best_match_units_21[u2] == -1** with the meaning the unit 2 do not match anything.

What about calling false positive a unit with less then 0.1 accuracy? If, for example, the tested sorting has less units then the GT sorting, all units will be matched, maybe with a 0.002 accuracy...A small accuracy could possibly be just random

spiketoolkit/study/groundtruthstudy.py
samuelgarcia: This is class method to setup:
 
So the first ttime to make the folrder:
  * study = GroundTruthStudy.setup(...)

And then to read/compute/compare:
  * study = GroundTruthStudy(folder)

spiketoolkit/comparison/groundtruthcomparison.py
samuelgarcia: I am not sure if we must split in 3.
I propose make a quite and dirty fix for the paper now and open a PR for more discussion about: FPunit/bad/redundant/intermediateFP...

samuelgarcia:
> What about calling false positive a unit with less then 0.1 accuracy? If, for example, the tested sorting > has  less units then the GT sorting, all units will be matched, maybe with a 0.002 accuracy...A small  >accuracy  could possibly be just random

accuracy is even another case, no ?
Here the main idea is does it match both way or not or no match (atleast in my mind).

alejoe91:
> > What about calling false positive a unit with less then 0.1 accuracy? If, for example, the tested sorting > has  less units then the GT sorting, all units will be matched, maybe with a 0.002 accuracy...A small  >accuracy  could possibly be just random
> 
> accuracy is even another case, no ?
> Here the main idea is does it match both way or not or no match (atleast in my mind).

That would be a redundant unit, no? I think that below 0.1 (or 0.05) one could say "this is a wrong unit" --> false positive

samuelgarcia:
The thing is that accuracy is not the agreement score and the matching is done on agreement score, no ?

alejoe91:
> The thing is that accuracy is not the agreement score and the matching is done on agreement score, no ?

Ok, we can use the agreement then (as in the reduntand units). I agreement is below 0.1 or 0.2 then it's a false positive. Of course for unmatched units we sould not have the accuracy.

samuelgarcia:
OK lets do that 
false positive : "best match but agreement<0.1 OR no match (-1)"

But, we still will have num_bad != (num_false_positive_unit + num_redundant).
Unit 2  with agreement between in range 0.1 - 0.5, will not be counted, no ?

alejoe91:
yes. So maybe we can only set 1 threshold ~ 0.2

< 0.2 or -1 --> false_positive
> 0.2 and not best match --> redundant

samuelgarcia:
Good idea. The problem is that treshold are inside each method so we can not garanty that they are the same for 2 succesive call.

alejoe91:
we can have an extra arg at instantiation! OK? `redundant_false_positive_threshold_agreement`

samuelgarcia:
threhold at instatiation are less flexible...

alejoe91:
One can overwrite it with the call. I think we need a compromize here..

alejoe91:
@samuelgarcia I would make the `setup` a normal method. So the workflow would be:

```
study_folder = 'some-folder'
study = GroundTruthStudy(study_folder)
study.setup(gt_dict)
study.run_sorters(sorter_list, mode='keep', engine='loop')
```

It looks cleaner and easier to me than having the static method..

samuelgarcia:
A bit cleaner but in my mind is ```study = GroundTruthStudy(study_folder)``` to open something already setup.

alejoe91:
maybe we could have an `open` function? to me `setup` sounds like making a new one

samuelgarcia:
@alejoe91 : I am totally stupid.

I have started spiekcomparison [here](https://github.com/SpikeInterface/spikecomparison/pull/1)

samuelgarcia:
I need to copy/paste all your changes...

samuelgarcia:
setup IS making a new one.

samuelgarcia:
study = GroundTruthStudy.setup() >>>> create

study = GroundTruthStudy(study_folder) >>>> open

maybe setupcan be change to create

alejoe91:
eheheh
wait to megre this PR. Then you can port the whole thing! :)

alejoe91:
I like `create`

samuelgarcia:
As a class method ?

alejoe91:
yes

alejoe91:
we can make the `open` later in `spikecomparison`

spiketoolkit/study/groundtruthstudy.py
```diff
@@ -8,131 +8,145 @@
...
-    
+
     def scan_folder(self):
         self.rec_names = get_rec_names(self.study_folder)
-        
         # scan computed names
-        self.computed_names = list(iter_computed_names(self.study_folder))  # list of pair (rec_name,sorter_name)
+        self.computed_names = list(iter_computed_names(self.study_folder))  # list of pair (rec_name, sorter_name)
         self.sorter_names = np.unique([e for _, e in iter_computed_names(self.study_folder)]).tolist()
+        self._is_scanned = True
```
samuelgarcia: self.scan_folder() is in init so 
self._is_scanned is always True.
So do we need this ?

samuelgarcia:
@alejoe91 : can I merge this now ?
So I can copy/paste changes directly in spikecomparison ?

spiketoolkit/study/groundtruthstudy.py
alejoe91: no we don't need it anymore.

The use is:
- new study
```
study = sc.GroundTruthStudy.create(study_folder, gt_dict)
```

- old study (TODO):
```
study = sc.GroundTruthStudy.open(study_folder)
```

both class functions that create the object. Should we move the discussion to spikecomparison?

alejoe91:
Yes! We will remove it later anyways

PR #207, opened 2019-07-15
Big rm because:

samuelgarcia:
Big rm because:
  * study+comparison go to spikecomparison
  * sorters go to spikesorters


alejoe91:
looks goos!

PR #221, opened 2019-08-27
(WIP) Test files

alejoe91:


colehurwitz:
Looks good. :) Should we just add to this pull request over time?

alejoe91:
Yep!

alejoe91:
@colehurwitz Could you make the missing test for the `normalize_by_quantile`? or @martinosorb ?
I don't understand how it should work properly.

colehurwitz:
It would be best if @martinosorb or @mhhennig did it as I don't understand how it works either.

alejoe91:
Hei @colehurwitz I would merge this. 
Only `test_normalize_by_quantile` in preprocessing and `test_compute_pca` are missing. For the latter, we can create fake waveforms starting from gaussians in the PC space. This approach could be used also for spikemetrics.

PR #224, opened 2019-08-30
save qc as properties and code cleanup

alejoe91:
@colehurwitz this is what I meant! :)

When you compute a metric, you can save the metric as unit property (default)


colehurwitz:
Looks good. We can raise an issue about epochs and unit properties in spikeextractors or spiketoolkit.

PR #225, opened 2019-09-02
Added verbose option

alejoe91:
- remove printProgressBar by default
- in compute_isi: if sampling_frequency is none, get it from sorting


colehurwitz:
The metric_calculator does the sampling frequency thing already.

PR #229, opened 2019-09-03
allow for instantiation of recording in metric calculator init

alejoe91:
@colehurwitz 

this allows to register the recording_extractor at the __init__ of MetricCalculator

### DIFF
--- a/spiketoolkit/validation/metric_calculator.py
+++ b/spiketoolkit/validation/metric_calculator.py
@@ -1,4 +1,5 @@
 import numpy as np
+from spikeextractors import RecordingExtractor, SortingExtractor
 import spiketoolkit as st
 import spikemetrics.metrics as metrics
 from spikemetrics.utils import Epoch, printProgressBar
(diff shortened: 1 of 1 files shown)


spiketoolkit/validation/metric_calculator.py
```diff
@@ -881,9 +887,9 @@ def get_metrics_df(self):
                 unit_ids = self.get_unit_ids()
                 for unit_id in all_unit_ids:
                     if unit_id in unit_ids:
-                        all_metrics[unit_id + len(all_unit_ids) * i] = metric[unit_ids.index(unit_id)]
+                        all_metrics[unit_ids.index(unit_id) + len(all_unit_ids) * i] = metric[unit_ids.index(unit_id)]
                     else:
-                        all_metrics[unit_id + len(all_unit_ids) * i] = ''
+                        all_metrics[unit_ids.index(unit_id) + len(all_unit_ids) * i] = ''
```
alejoe91: this is a fix when unit_ids don't start fom 0

colehurwitz:
I’m not sure this is a good change or not. The recording extractor is already easy to set using the set_recording function and, in isolation, the recording extractor only allows for the user to run SNR

colehurwitz:
You still need to compute quality metric data to do anything other than SNR. I separated relevant data preprocessing for that reason.

colehurwitz:
What convenience does this add? Do you want to compute SNR immediately?

alejoe91:
I think it does no harm. It removes the need of calling the set_recording.

colehurwitz:
Sounds good.

PR #231, opened 2019-09-03
fix snr when recording is short

alejoe91:


colehurwitz:
Isn't this addressed in the if and else statement below your change?

alejoe91:
no it fails!

alejoe91:
ahhhh
sorry I missed it!!!! you are right, it's probably somewhere else..sorry!

colehurwitz:
Maybe don't merge until we discuss a little for some of these metric things

alejoe91:
I see, it's here: 
`randint(0, recording.get_num_frames() - n_frames)`

we are trying to fix the build in the docs. We'll release a 0.5 when it's fixed

alejoe91:
I think it shuld be: `n_frames >= recording.get_num_frames():` instead of `n_frames > recording.get_num_frames():`

alejoe91:
exactly, it fails when the duration is exactly the same. Will revert the changes and make a new PR

PR #233, opened 2019-09-06
updated spiketoolkit for shared functions

colehurwitz:


colehurwitz:
I will disable matplotlib functionality in spikeglx as it isn't used

samuelgarcia:
All example should be remove from here they are replaced by spikeinterface/examples

samuelgarcia:
Same here.

colehurwitz:
Do examples need to be removed from all repos? My pull request doesn't cover this. Maybe you can make an issue?

samuelgarcia:
I think yes.
@alejoe91 ?

The good thing with example in spikeinterface is that they are generated (and centralized).

colehurwitz:
Removed them all

PR #243, opened 2019-09-19
added 'remove_empty_units' flag to export_to_phy funciton

Shawn-Guo-CN:
add a 'remove_empty_units' flag to function **export_to_phy**, in order to remove empty units that may appear in input sortings.

### DIFF
--- a/spiketoolkit/postprocessing/postprocessing_tools.py
+++ b/spiketoolkit/postprocessing/postprocessing_tools.py
@@ -682,7 +682,7 @@ def export_to_phy(recording, sorting, output_folder, nPC=3, electrode_dimensions
                   grouping_property=None, ms_before=1., ms_after=2., dtype=None, amp_method='absolute', amp_peak='both',
                   amp_frames_before=3, amp_frames_after=3, max_num_waveforms=np.inf, max_num_pca_waveforms=np.inf,
                   recompute_waveform_info=True, save_features_props=False, write_waveforms=False, verbose=False,
-                  seed=0, remove_empty_units=True):
+                  seed=0):
     '''
     Exports paired recording and sorting extractors to phy template-gui format.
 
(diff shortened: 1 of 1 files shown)


colehurwitz:
Looks good! One question, why would the user ever set this flag to false? Would that throw an error?

Shawn-Guo-CN:
> Looks good! One question, why would the user ever set this flag to false? Would that throw an error?

Just in case that the user deeply believe that there shouldn't be any empty unit in their data. Of course, I'm not sure whether this assumption is meaningful in real application scenarios.

colehurwitz:
I think export to phy should just automatically get rid of empty units. I don’t think we need a flag for it. I would start with that.

spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -733,8 +733,18 @@ def export_to_phy(recording, sorting, output_folder, nPC=3, electrode_dimensions
...
+        _spikes = sorting.get_unit_spike_train(unit_id)
+        if _spikes.shape[0] == 0: _empty_flag = True
+    if _empty_flag:
+        print('Warning: empty units have been removed when being exported to Phy')
+        sorting = st.curation.threshold_num_spikes(sorting, 1)
+
     if len(sorting.get_unit_ids()) == 0:
-        raise Exception("No units in the sorting result, can't save to phy.")
+        raise Exception("No non-empty units in the sorting result, can't save to phy.")
+
```
samuelgarcia: Here is another problem : no units at all. Please come back to the previsous Exception.

samuelgarcia:
Hi.
thank you for this.
My idea was to add an option in the function signautre.
If you clean anyway, the first list is useless.
I think.

Shawn-Guo-CN:
Now, what the function export_to_phy does is to check:

1. whether input recording and sorting has correct type;
2. whether there exists empty unit, remove them if so;
3. whether there are units in soring.

At the current time, it seems solid to export sortings to phy.

PR #246, opened 2019-09-22
Exclude units without spikes from all calculations (still a bug here)

mhhennig:
This code still breaks when `unit_ids` is supplied. The trouble is that the metric_data is computed for all units in an extractor, regardless of `unit_ids`.

### DIFF
--- a/spiketoolkit/validation/metric_calculator.py
+++ b/spiketoolkit/validation/metric_calculator.py
@@ -43,15 +43,20 @@ class MetricCalculator:
 
         if unit_ids is None:
             unit_ids = sorting.get_unit_ids()
-            
+        
         # only use units with spikes to avoid breaking metric calculation
-        num_spikes_per_unit = [len(sorting.get_unit_spike_train(u)) for u in unit_ids]
+        num_spikes_per_unit = [len(sorting.get_unit_spike_train(unit_id)) for unit_id in sorting.get_unit_ids()]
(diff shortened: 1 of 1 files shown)


colehurwitz:
How I will fix this is to make the `unit_ids = sorting.get_unit_ids()` only when `unit_ids == None`. When unit_ids is a list of unit ids, I would do:  `unit_ids = intersection(unit_ids, sorting.get_unit_ids())`

spiketoolkit/validation/metric_calculator.py
```diff
@@ -42,6 +43,15 @@ def __init__(self, sorting, recording=None, sampling_frequency=None, unit_ids=No
...
             unit_ids = sorting.get_unit_ids()
+            
+        # only use units with spikes to avoid breaking metric calculation
+        num_spikes_per_unit = [len(sorting.get_unit_spike_train(u)) for u in unit_ids]
+        sorting = ThresholdCurator(sorting=sorting, metrics_epoch=num_spikes_per_unit)
+        sorting.threshold_sorting(0, 'less_or_equal')
+
+        unit_ids = sorting.get_unit_ids()
+        if len(unit_ids)==0:
+            raise ValueError("No spikes found.")
```
samuelgarcia: I would prefer "no units found"

colehurwitz: Fixed in new commits

colehurwitz:
Ready to merge

Issue #249, opened 2019-09-23
Performance of filtering in FIlterRecording is much lower than direct filtering using scipy.signal 


teristam:
Consider the following test script:
```
import spikeinterface as si
import spikeinterface.extractors as se
import spikeinterface.toolkit as st
from scipy.signal import butter,filtfilt
import numpy as np
import time


Fs = 30000
signal = np.random.rand(16,120*Fs)
recording = se.NumpyRecordingExtractor(signal,Fs)

def filterRecording(recording, sampling_freq, lp_freq=300,hp_freq=6000,order=3):
    fn = sampling_freq / 2.
    band = np.array([lp_freq, hp_freq]) / fn

    b, a = butter(order, band, btype='bandpass')

    if not (np.all(np.abs(np.roots(a)) < 1) and np.all(np.abs(np.roots(a)) < 1)):
        raise ValueError('Filter is not stable')
    
    for i in range(recording._timeseries.shape[0]):
        recording._timeseries[i,:] = filtfilt(b,a,recording._timeseries[i,:])

    return recording


# Direct filtering with scipy
start = time.time()
filterRecording(recording,Fs) #for faster operation later
print(f'\n Scipy.filtfilt lasped time: {time.time()-start} ')


# Filtering using the filter object
start = time.time()
recording_sort = st.preprocessing.bandpass_filter(recording, freq_min=300, freq_max=6000,cache=True, type='butter')
print(f'\n st.preprocessing.bandpass_filter lasped time: {time.time()-start} ')

# Filtering using the filter object (small chunk size)
start = time.time()
recording_sort = st.preprocessing.bandpass_filter(recording, freq_min=300, freq_max=6000,cache=True, type='butter',chunk_size=1000)
print(f'\n st.preprocessing.bandpass_filter (small chunk) lasped time: {time.time()-start} ')
```

On my macbook, the output is
```
 Scipy.filtfilt lasped time: 1.3340470790863037 
 st.preprocessing.bandpass_filter lasped time: 2.848065137863159 
 st.preprocessing.bandpass_filter (small chunk) lasped time: 16.753363847732544 
```
In the case where there are lots of spikes, the filtering overhead of the small chunk size option will definitely overshadow its benefit.

alejoe91:
[...]

Issue #248, opened 2019-09-23
compute_all_metric_data is slow on large dataset


teristam:
When calling `compute_all_metric_data`, in `postprocessing_tools._get_quality_metric_data`, `get_unit_amplitudes` will return the amplitude of all spike waveforms, which can be very slow for dataset with many spikes. There should be an option to specify a subset of the waveform to return. Usually it may not necessary to compute the metrics on all spikes.

alejoe91:
@teristam good point! We are actually updating the metric calculator on these days to fix these issues (plus others!) Will reference to the PR once it's open. @colehurwitz we should reduce the default number of spikes in all metrics

colehurwitz:
I think the challenge is that Josh's code expects all input arrays to have the same length as all spikes in the recording. See https://github.com/SpikeInterface/spikemetrics/blob/master/spikemetrics/metrics.py

I am not sure how to overcome this constraint.

colehurwitz:
We could do the epochs **before** the metrics are calculated rather than during. Josh's code takes in all the data and then does epochs.

colehurwitz:
Could we parallelize some of the post processing tools to make all of this faster?

samuelgarcia:
Hi @colehurwitz : I don't get this epoch stuff, sorry I did not have time to read the code.
I think we should hack the original code by adding **selected_spike_index** (or something like that) to reduce the number of spikes everywhere.
Ideally this  **selected_spike_index** would be one randomly once with a seed to ensure reproducibility.

colehurwitz:
Also, we maybe make a new function for computing all of the quality metric data except for amplitudes. Amplitudes are only used in the cutoff metric.

colehurwitz:
I think sub-selecting spikes might cause issues, Samuel. It would definitely bias ISI violations and presence ratios right?

samuelgarcia:
Totally, sorry. I had in mind waveform mostly.

alejoe91:
I think that for isi and presence using all spikes it's no problem as only spiketrains are used. The problem is when metrics that require wf and templates are used. In that case it's safe to use a subsample of the spikes as we compute the templates as the median

colehurwitz:
I think it is a bit challenging to change josh’s code for this.. maybe we can get Josh to work on that.

colehurwitz:
Do you think postprocessing tools is parallelizable as well @alijoe91?

alejoe91:
[...]

PR #251, opened 2019-09-24
Improved filtering

alejoe91:
- moved out the butter filter creation in the init
- cache option now filters all the data and dumps it into a bin recording

Should fix:
https://github.com/SpikeInterface/spiketoolkit/issues/249  https://github.com/SpikeInterface/spiketoolkit/issues/248

I tested the performance on 120s 16 channels:

**Default filter**
```
rec_f = st.preprocessing.bandpass_filter(rec)

# filter all traces
t_start = time.time()
rec_f.get_traces()
print(time.time() - t_start)

2.017151117324829 s

# get waveform (before internal caching)
t_start = time.time()
wf1 = st.postprocessing.get_unit_waveforms(rec_f, sort)
print(time.time() - t_start)

2.023608446121216 s
```

**Cache** - the filtering is performed at initialization and dumped to file (a bit slower)
```
# initialization + filter + save
t_start = time.time()
rec_c = st.preprocessing.bandpass_filter(rec, cache=True)
print(time.time() - t_start)

3.8165152072906494 s

# get waveform 
t_start = time.time()
wf1 = st.postprocessing.get_unit_waveforms(rec_c, sort)
print(time.time() - t_start)

0.08617234230041504 s
```

@samuelgarcia @teristam @magland 
I think this will speed up **A LOT** the waveform retrieval. If you agree to this design I will propagate the cache option to all other preprocessors.

### DIFF
--- a/spiketoolkit/preprocessing/bandpass_filter.py
+++ b/spiketoolkit/preprocessing/bandpass_filter.py
@@ -1,6 +1,8 @@
 from .filterrecording import FilterRecording
 import numpy as np
 from scipy import special
+import tempfile
+import spikeextractors as se
 from scipy.signal import butter, filtfilt
 
 try:
@@ -28,17 +30,28 @@ class BandpassFilterRecording(FilterRecording):
     ]
     installation_mesg = "To use the BandpassFilterRecording, install scipy: \n\n pip install scipy\n\n"  # err
 
-    def __init__(self, recording, freq_min=300, freq_max=6000, freq_wid=1000, type='fft', order=3, chunk_size=30000, cache=False):
+    def __init__(self, recording, freq_min=300, freq_max=6000, freq_wid=1000, type='fft', order=3,
+                 chunk_size=30000, cache=False):
         assert HAVE_BFR, "To use the BandpassFilterRecording, install scipy: \n\n pip install scipy\n\n"
         self._freq_min = freq_min
         self._freq_max = freq_max
         self._freq_wid = freq_wid
         self._type = type
         self._order = order
+        self._chunk_size = chunk_size
+
+        if self._type == 'butter':
+            fn = recording.get_sampling_frequency() / 2.
+            band = np.array([self._freq_min, self._freq_max]) / fn
+
+            self._b, self._a = butter(self._order, band, btype='bandpass')
+
+            if not np.all(np.abs(np.roots(self._a)) < 1):
+                raise ValueError('Filter is not stable')
+        if cache:
+            self._chunk_size = None
         FilterRecording.__init__(self, recording=recording, chunk_size=chunk_size, cache=cache)
         self.copy_channel_properties(recording)
-        if cache:
-            self._traces = self.get_traces()
 
     def filter_chunk(self, *, start_frame, end_frame):
         padding = 3000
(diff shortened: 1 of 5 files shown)


colehurwitz:
Wow!! This is awesome

colehurwitz:
Can this be implemented in the filter base class or does it have to be for each filter specifically?

samuelgarcia:
@teristam : Yes there is a **delete_output_folder** option False by default.

colehurwitz:
> @teristam : Yes there is a **delete_output_folder** option False by default.

I think that @teristam is referring to deleting the cached file when filtering not for when sorting.

samuelgarcia:
Sorry.
I am just reading the code.
I was refering to another question.

@alejoe91 I am not sure to be happy that the trick of cache is done in the function and not in the classs itself.
Furthermore it could be done at a baseclass level like a mixin class **ProcessorCache**.

I will read it more carfully later.

alejoe91:
I agree. This was just an example to test performance. @magland suggested an extractor called Cache recording extractor that can be initialized by passing a recording extractor. I think it's a good idea

alejoe91:
@samuelgarcia I think that the function calls will still be the right place to wrap the preprocessor extractors in the cache recording object. I'll give it a try!

alejoe91:
@teristam tested now on 7000 waveforms...

Without caching: 96 s (without running a `recording.get_traces()` before, as one would normally do)
With new caching: 0.67 s

alejoe91:
When https://github.com/SpikeInterface/spikeextractors/pull/264 is merged tests should pass!

Let me know what you guys think!

teristam:
@alejoe91 that's awesome! Just curious how would a disk-cached signal be used in the various sorters? Will the entire signal be loaded into memory?

spiketoolkit/preprocessing/bandpass_filter.py
```diff
@@ -1,6 +1,8 @@
 from .filterrecording import FilterRecording
 import numpy as np
 from scipy import special
+import tempfile
```
samuelgarcia: not needed anymore

samuelgarcia:
Hi Alessio,
this looks good some questions:
  * why for FFT case the kernel is computed for each chunk ?
  * what append if we do this:
```python

rec0 = se.SomeExtrcator
rec_cache = bandpass_filter(rec0, freq_min=3000, freq_max=6000,
    type='butter', order=3, cache=True)
some_trace = rec_cache.get_traces(0, 1000)
del rec_cache
```
My guess is that the del will fail because some_trace still memmap the file that cannot be removed.

What do you think ?

alejoe91:
> @alejoe91 that's awesome! Just curious how would a disk-cached signal be used in the various sorters? Will the entire signal be loaded into memory?
> 
> A typical workflow using the package will be
> 
> 1. Load the experimental data from disk
> 2. Make a RecordingExtractor
> 3. Do filtering
> 4. Do sorting
> 5. Calculate some metrics for curation
> 6. Save the sorting results
> 
> I am wondering if we save the filtered signals to disk, but later load it back to memory in sorting, does it offer any benefit than just keeping the entire filtered signal in memory?

@teristam 1. and 2. are the same: loading a recording creates a RecordingExtractor. 
Usually CLI-based sorters (klusta and SC) and matlab based (kilosort, ironclust, waveclus) make another copy of the signals in the output folder. However MS4 will benefit a lot from caching as it uses the recrding extractor directly!

alejoe91:
@samuelgarcia @colehurwitz I think this is ready to merge on my side

colehurwitz:
As long as @samuelgarcia is ok with it, I am ok with it :D

samuelgarcia:
@alejoe91 : you did not answer to my comment from yesterday but it is totally OK for me.
:)

alejoe91:
> Hi Alessio,
> this looks good some questions:
> 
> * why for FFT case the kernel is computed for each chunk ?
> * what append if we do this:
> 
> ```python
> rec0 = se.SomeExtrcator
> rec_cache = bandpass_filter(rec0, freq_min=3000, freq_max=6000,
>     type='butter', order=3, cache=True)
> some_trace = rec_cache.get_traces(0, 1000)
> del rec_cache
> ```
> 
> My guess is that the del will fail because some_trace still memmap the file that cannot be removed.
> 
> What do you think ?

Saw it now! I can test it tomorrow. Let's keep this open for the moment?

alejoe91:
@samuelgarcia the kernel depends on the chunk size and that could change depending on padding. If you filter a range that starts from zero or at the end you will end up with a different kernel.

I tried what you suggested and it works with no problems. So maybe the traces are not memmaped because of this:
https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extraction_tools.py#L245
The destructor is called and the file is deleted...

I also tried to use `np.memmap.transpose` instead and it still works. Strange. Should I merge anyways? I don't think users would delete the object and continue to do stuff.

samuelgarcia:
So lets merge.
Stange that there is not bug...
Normally caching with memmap can lead lead to boudaries effect.
I guess under window it will be differents.
We will se later if we have problems, this cache is indeed a very good thing and the implementation is super simple, I like it.

PR #254, opened 2019-09-28
Fix metrics when max_num_spikes and pca is less than inf

alejoe91:
This allows to run metrics with a few number of spikes

- postprocessing tools can return indexes of spikes used to compute waveforms, amplitudes, and pc scores
- the indexes are used to extract two extra set otf spike_clusters and spike_times: spike_times_pca, spike_times_amps
- metrics using amplitudes or pca use this subsets of spike times (to get epochs idxs) and spike clusters and pass them to spikemetrics functions

### DIFF
--- a/spiketoolkit/curation/curationsortingextractor.py
+++ b/spiketoolkit/curation/curationsortingextractor.py
@@ -41,12 +41,12 @@ class CurationSortingExtractor(SortingExtractor):
         valid_unit_id = False
         spike_train = np.asarray([])
         for root in self._roots:
-            if (root.unit_id == unit_id):
+            if root.unit_id == unit_id:
                 valid_unit_id = True
                 full_spike_train = root.get_spike_train()
                 inds = np.where((start_frame <= full_spike_train) & (full_spike_train < end_frame))
                 spike_train = full_spike_train[inds]
-        if (valid_unit_id):
+        if valid_unit_id:
             return spike_train
         else:
             raise ValueError(str(unit_id) + " is an invalid unit id")
@@ -66,7 +66,7 @@ class CurationSortingExtractor(SortingExtractor):
         for i in range(len(self._roots)):
             root_id = self._roots[i].unit_id
             root_ids.append(root_id)
-        if (unit_id in root_ids):
+        if unit_id in root_ids:
             root_index = root_ids.index(unit_id)
             print(self._roots[root_index])
         else:
@@ -80,14 +80,14 @@ class CurationSortingExtractor(SortingExtractor):
         unit_ids: list
             The unit ids to be excluded
         '''
-        if (len(unit_ids) == 0):
+        if len(unit_ids) == 0:
             return
         root_ids = []
         for i in range(len(self._roots)):
             root_id = self._roots[i].unit_id
             root_ids.append(root_id)
 
-        if (set(unit_ids).issubset(set(root_ids)) and len(unit_ids) > 0):
+        if set(unit_ids).issubset(set(root_ids)) and len(unit_ids) > 0:
             indices_to_be_deleted = []
             for unit_id in unit_ids:
                 root_index = root_ids.index(unit_id)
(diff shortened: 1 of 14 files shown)


alejoe91:
@mhhennig can you try this? It should fix and speed up the metrics extraction

colehurwitz:
You are a saint... Btw, in the quality_metric functions, can you switch the compute_metric_data calls to the correct one now (pca, amplitudes, etc.)?

alejoe91:
You mean in the assertion? The pca and amplitude function calls should be in place

colehurwitz:
So for example, in `compute_d_primes` in `quality_metrics` (the functions), it still calls `compute_all_metric_data` rather than `get_pca_metric_data`

alejoe91:
Yup I'll fix those

colehurwitz:
Parentheses are my vice

spiketoolkit/curation/curationsortingextractor.py
```diff
@@ -81,21 +80,21 @@ def exclude_units(self, unit_ids):
...
-        if(len(unit_ids) == 0):
+        if len(unit_ids) == 0:
             return
         root_ids = []
         for i in range(len(self._roots)):
             root_id = self._roots[i].unit_id
             root_ids.append(root_id)
 
-        if(set(unit_ids).issubset(set(root_ids)) and len(unit_ids) > 0):
+        if set(unit_ids).issubset(set(root_ids)) and len(unit_ids) > 0:
```
samuelgarcia: Is "set" fatser than np.in1d ?
I personnally find in1d more convinient to read.

spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -11,7 +11,7 @@
 def get_unit_waveforms(recording, sorting, unit_ids=None, grouping_property=None, channel_ids=None,
                        ms_before=3., ms_after=3., dtype=None, max_spikes_per_unit=np.inf,
                        save_as_features=True, compute_property_from_recording=False, verbose=False,
-                       seed=0):
+                       seed=0, return_idxs=False):
```
samuelgarcia: I think for performence reason both "return_idxs" and "waveforms" should be numpy arrays preallocated at the good lenght.

spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -848,66 +892,205 @@ def _get_random_spike_waveforms(recording, sorting, unit, max_spikes_per_unit, s
...
+    spike_clusters = np.array([])
+    amplitudes = np.array([])
+
+    for i_u, id in enumerate(sorting.get_unit_ids()):
+        st = sorting.get_unit_spike_train(id)
+        amp = amplitudes_list[i_u]
+
+        if len(amp) < len(st):
+            cl = [i_u] * len(amp)
+            spike_times = np.concatenate((spike_times, st[amp_idxs[i_u]]))
```
samuelgarcia: concatenating in a loop is a bad pratice. It is better to preallocate the good array sioze from the beggning or concatenating at the end a big list.

samuelgarcia:
Hi Alessio.
This is totally amazing.
I have small query : for such big changes I would be more efficient to make only logical changes.
Here there is a mix between a super work and esthetic changes due to linter.
This make the review quite slow.

alejoe91:
sorry @samuelgarcia you are right! It's @colehurwitz's fault!!!

spiketoolkit/postprocessing/postprocessing_tools.py
alejoe91: I could fix that here since we now the indexes, but I think that since these are 1d array it's fast anyways

spiketoolkit/postprocessing/postprocessing_tools.py
alejoe91: waveforms and indexes can have differet shapes. Even if max_spikes_per_unit=300, if a spike train has less than 300 spikes than we can't pre-allocate memory..I'm afraid we have tu use lists

spiketoolkit/curation/curationsortingextractor.py
alejoe91: @colehurwitz wrote this. I think this i quite fast anyways..

samuelgarcia:
:)
So easy to accuse your camarde.

samuelgarcia:
OK for me.

PR #266, opened 2019-10-21
No need to save waveforms to phy

alejoe91:
There is no need to save waveforms when exporting to phy (this was form an old attempt)


samuelgarcia:
Can we merge ?

alejoe91:
Yep!!

alejoe91:
Also the other PR

Issue #265, opened 2019-10-18
Common_reference doesn't work for single channels that are not in a list [fixed]


maartensc:
When running the preprocessing function common_reference, with referencing to a single channel, I got the error:

spikeextractors/spikeextractors/extractors/bindatrecordingextractor/bindatrecordingextractor.py, line 64, in get_traces,
channel_ids = [self._channels.index(ch) for ch in channel_ids]
TypeError: 'numpy.int64' object is not iterable

This is easily fixed by changing line 37 to `self._ref_channel = [ref_channels]`, so channel_ids always becomes a list, and thus becomes iterable.

Maybe I was just doing something wrong in the first place, but this easily fixed it for me, so maybe if others also experience this error, it could be a solution to change it in the code.

alejoe91:
@maartensc thanks for spotting this. Indeed it should work with an int as well. I'll fix that!

PR #267, opened 2019-10-21
Fix common reference when ref_channels is int

alejoe91:
This ensures that ref_channels is always a list when calling `get_traces`. It fixes https://github.com/SpikeInterface/spiketoolkit/issues/265


samuelgarcia:
Can we merge ?

colehurwitz:
I think we can merge this, right?

alejoe91:
Yes yes

PR #279, opened 2019-11-28
removed returning single list when unit ids is 1

colehurwitz:
We had the feature of returning a single list (instead of a list of list) when the number of units was 1. I removed this since it causes issues for downstream functions. Had to adjust a bunch of code to make it work.

@alejoe91 can you check elsewhere to see if this may cause problems? No problems in spiketoolkit I think, but not sure about widgets.


colehurwitz:
This is for both postprocessing (waveforms, templates, etc.) and quality metrics (snrs, d-primes, etc.).

PR #292, opened 2020-01-29
sortingcomponents module

alejoe91:
Added sortingtools module with `detect_spikes` function

https://github.com/SpikeInterface/spikesorters/issues/75


alejoe91:
@magland @samuelgarcia @colehurwitz something like this?

colehurwitz:
Looks about right! Maybe we can call the module `sortingcomponents` and then have files like: `sortingcomponents/detection.py`, `sortingcomponents/clustering.py`, etc. Since we may add more.

alejoe91:
@colehurwitz better?

colehurwitz:
Yep!! Nice work. This is a cool addition.

alejoe91:
Can I merge this?

samuelgarcia:
OK for me.
I like the place where it is implmenented!






 (and don't like so much the idea of detecting spikes independently by channel. But it is a good start)

PR #294, opened 2020-02-15
Quality metric refactor [WIP]

colehurwitz:
I still need to update documentation and add more tests, but I finished the refactorization of quality metrics that I discussed with @alejoe91 . Apologies to @samuelgarcia that I committing everything at once instead of doing multiple commits over time (I know that makes it hard to review). I kept the API the same mostly so that current users won't feel the difference.

Summary of refactorization:

Major
1. **Added**: a parent class QualityMetric
2. **Added**: a subclass for each quality metric (e.g. FiringRate, DPrime, etc.) that can **compute** a metric or **threshold** a sorting extractor with the metric. Each class now stores the parameters that are used for computing the metric (the functions access these params). GUI params are cleaned and added to these classes as well for curation. **Stored** in the quality_metric_classes folder.
3. **Added**: a data object, MetricData, which is used to compute data for quality metrics.
4. **Added**: a parameter_dictionaries file that contains dictionaries of shared parameters for data computation.
5. **Removed**: MetricCalculator
6. **Removed**: curationlist
7. **Added**: validationlist
8. **Reimplemented**: all metrics and threshold metrics with the new refactoring.
9. **Tested** and verified that all the new functions have equivalent return values to old methods.

Small
1. **Fixed** save properties for quality metrics (they were not being saved correctly).
2. **Removed** checking for 0 spike units (this was causing saving issues).
3. **Changed** names of thresholders to be plural for all metrics.
4. **Added**: utils folder in quality_metric_classes that contains thresholdcurator, curationsortingextractor, and validation_tools.

TODO upon approval:
1. Update all documentation.
2. Improve testing.
3. Enforce pep8
4. Make spikely compliant with new refactorization.


samuelgarcia:
Hi Cole.
This looks like a big refactoring! Lot of works, congrats.
I will try to have a look quite soon.
Sam

colehurwitz:
Thanks @samuelgarcia ! I had to make sure it was still compatible with the GUI so that was tricky. I tried my best not to have any duplicate default parameters anywhere so I used the parameter_dictionaries.

colehurwitz:
Could we take a look at this today? @alejoe91 @samuelgarcia . If we merge it, I can update documentation and the GUI. I need to go through all GUI params in spikeextractors/spiketoolkit and check them. Then I need to test the GUI.

samuelgarcia:
Hi cole.
Sorry I did not find time to review this yet.
Alessio in quarantine should do it soon.

I will be in holiday tomorow evening. I hope I will have time to do it.

At the moment I have only comments:
When you do this
```python
def one_function(...,
    epoch_params=get_epoch_params(),
    verbose=NumSpikes.params['verbose'],
):
```
be aware that default for this function will NOT be dynamic because the signature is done once.

In short, if **get_epoch_params** change or **NumSpikes.params['verbose']** change the default kargs of the **one_function** will not change, they will remain the same at the beginning until python restarted.

the signature of a function is executed only one in python. so get_epoch_params() is call once only.

Is it the behavior you want ?


Sam

colehurwitz:
@samuelgarcia Thanks for letting me know about this behaviour. I think it is ok since I don't expect users to changes the defaults used in our codebase. I just expect them to input in new parameter dictionaries that contain their specific values.

spiketoolkit/validation/validation_list.py
```diff
@@ -0,0 +1,63 @@
+from .quality_metric_classes.metric_data import MetricData
```
alejoe91: I would remove the classes from the init, it makes things more confusing for the user IMO

spiketoolkit/validation/validation_list.py
```diff
@@ -0,0 +1,63 @@
...
+from .quality_metric_classes.parameter_dictionaries import (
+    get_recording_params,
+    get_amplitude_params,
+    get_pca_scores_params,
+    get_epoch_params,
+    get_feature_params,
+    update_param_dicts,
+)
+
+curation_full_list = [
```
alejoe91: can you move the curation stuff in the curation module?

spiketoolkit/validation/quality_metric_classes/utils/thresholdcurator.py
```diff
@@ -1,6 +1,4 @@
 from .curationsortingextractor import CurationSortingExtractor
-import spiketoolkit as st
```
alejoe91: maybe the thresholdcurator and curationsortingextractor should be in the curation module?

spiketoolkit/validation/validation_list.py
colehurwitz: Yes, that is fair.

spiketoolkit/validation/validation_list.py
colehurwitz: The challenge with this is that the threshold functions are now in each QualityMetric class so I can't move it over to curation without creating a circular dependency. I am not sure how to deal with that.

spiketoolkit/validation/quality_metric_classes/utils/thresholdcurator.py
colehurwitz: Same issue as above.

spiketoolkit/validation/validation_list.py
colehurwitz: Validation depends on curation (since each quality metric thresholds)
Curation depends on validation (because it needs to compute quality metrics).

spiketoolkit/validation/validation_list.py
colehurwitz: Fixed

alejoe91:
@colehurwitz merge when you want :)

colehurwitz:
If @samuelgarcia has no objections, then I can merge!

PR #304, opened 2020-03-10
fix epochs

colehurwitz:
Setting all the infinities to None

### DIFF
--- a/spiketoolkit/validation/quality_metric_classes/amplitude_cutoff.py
+++ b/spiketoolkit/validation/quality_metric_classes/amplitude_cutoff.py
@@ -37,8 +37,14 @@ class AmplitudeCutoff(QualityMetric):
     def compute_metric(self, save_as_property):
         amplitude_cutoffs_epochs = []
         for epoch in self._metric_data._epochs:
+            start_frame = epoch[1]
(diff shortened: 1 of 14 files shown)


colehurwitz:
Should work with the spikeextractors pull request

samuelgarcia:
Hi COle,

Here I see some copy/paste and so I would make a function like this one:

```python
def get_in_epoch_bool_mask(spike_times, epoch):
    start_frame = epoch[1]
    end_frame = epoch[2]
    if start_frame is None:
        start_frame = 0
    if end_frame is None:
        end_frame = np.inf
    in_epoch = (spike_times>=start_frame) & (spike_times<=end_frame)
    return in_epoch
```

Also for limits I would use >= for low limit and < for high limit, no ?

colehurwitz:
Yes, I will make a function for this. I agree about the less than logic, but I need to check josh’s code to see which one he did.

samuelgarcia:
Perfect :)

colehurwitz:
Thanks for the suggestion! I will keep an eye out for redundant code in the future.

colehurwitz:
Should we release spikeextractors so that this will pass? @alejoe91

PR #308, opened 2020-03-26
Simplify validation and curation argument passing using kwargs

alejoe91:
Instead of passing several parameters for the `validation` and `curation` functions, now one can use `**kwargs`. To get a list of available `**kwargs`, we can refer to the docstrings or use the function `st.validation.get_kwargs_params()` or `st.curation.get_kwargs_params()`

### DIFF
--- a/spiketoolkit/validation/quality_metric_classes/parameter_dictionaries.py
+++ b/spiketoolkit/validation/quality_metric_classes/parameter_dictionaries.py
@@ -130,19 +130,19 @@ def update_param_dicts_with_kwargs(kwargs):
         for k in kwargs.keys():
             if k in recording_params.keys():
                 recording_params[k] = kwargs[k]
-    if np.any([k in amplitude_params.keys() for k in kwargs.keys()]):
+    elif np.any([k in amplitude_params.keys() for k in kwargs.keys()]):
         for k in kwargs.keys():
             if k in amplitude_params.keys():
                 amplitude_params[k] = kwargs[k]
-    if np.any([k in pca_scores_params.keys() for k in kwargs.keys()]):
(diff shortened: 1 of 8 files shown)


spiketoolkit/validation/quality_metric_classes/parameter_dictionaries.py
```diff
@@ -1,122 +1,150 @@
...
+    all_params.update(get_recording_params())
+    all_params.update(get_amplitude_params())
+    all_params.update(get_pca_scores_params())
+    all_params.update(get_epoch_params())
+    all_params.update(get_feature_params())
+
+    return all_params
+
+
+def update_param_dicts_with_kwargs(kwargs):
```
colehurwitz: I think your logic here can be simplified by a lot. Instead of checking if the keys are in each dict and having many loops, I would just loop over the kwargs once and update the dict it is in (there shouldn't be any kwargs in each dict)

```
for k in kwargs.keys():
      if k in recording_params.keys():

      elif k in amplitude_params.keys():

       .
       . 
       .
```

spiketoolkit/validation/quality_metric_classes/parameter_dictionaries.py
```diff
@@ -1,122 +1,150 @@
...
-            set(get_feature_params().keys())
-        ):
-            raise ValueError("Improper parameter entered into the feature param dict.")
-        else:
-            feature_params = OrderedDict(get_feature_params(), **feature_params)
-            param_dicts.append(feature_params)
-
-    return param_dicts
+
+def get_kwargs_params():
```
colehurwitz: This is great.

colehurwitz:
Looks good! Just some recommended changes (if they make sense) for the updating of the dicts.

PR #311, opened 2020-03-30
Use recording.is_filtered and avoid re-filtering if recordings are already filtered

alejoe91:


alejoe91:
needs this PR to pass https://github.com/SpikeInterface/spikeextractors/pull/342

Issue #306, opened 2020-03-25
default max_spikes_per_unit is inf


llobetv:
Hi guys, 

I try convert MS4 data to phy with your interface but I encounter a big issue, each time my memory was overload... 
I check export_to_phy function and it appear that default value for max_spikes_per_unit is set to infinite and their is no way to change it... 
Maybe you could add this variable in export_to_phy function?

Best regards,
Victor

alejoe91:
HI Victor,

For exporting to phy, you need to compute pca_scores for all waveforms, so that parameter cannot be changed.
Are you using the master branch? I recently added an option (true by default) that creates memmap objects for all waveforms, so they don't get loaded to memory. 

How long are your recordings and how many channels? 

Cheers
Alessio

mhhennig:
Also, could you show the code you use for this? I suspect the memmapping may not yet work in all cases...

llobetv:
Hi, 

I used dev branch of spikeInterface so it should be on date
I record 30mn with 64 channels (and some unit with 200000 events)

Here my code:

#####################################################################
from spikeinterface import extractors as se
from spikeinterface import toolkit as st

import numpy as np
import os

import pdb


verbose=True
#change tempdir of mountainSort
os.environ['TEMPDIR'] = "/mnt/hdd/mountainlab/mountainlab/"


file_path = "/mnt/ssd/spikeinterface_test/rec_200306_094252/amplifier_sorted.dat"
prb_file = "64_para.prb"

output_folder = "./MS_test/"

ms4_params= {'detect_sign': 0, 'adjacency_radius': 51, 'freq_min': 150, 
              'freq_max': 4000, 'filter': True, 'whiten': True, 'curation': False, 
              'num_workers': 40, 'clip_size': 40, 'detect_threshold': 3, 
              'detect_interval': 10, 'noise_overlap_threshold': 0.15}

numchan = 64
sampling_frequency = 30000
dtype = np.int16




#launch analysis
sorting_MS4 = ss.run_mountainsort4(recording=recording_prb, output_folder=output_folder, verbose=verbose, **ms4_params)

#export to phy
st.postprocessing.export_to_phy(recording, sorting_MS4, "phy/", n_comp=2, 
                                electrode_dimensions=None, grouping_property=None, 
                                ms_before=1., ms_after=1.5, dtype=None, 
                                amp_method='absolute', amp_peak='both', 
                                amp_frames_before=3, amp_frames_after=3, 
[...]

PR #313, opened 2020-03-30
Optional computation of pc features when exporting to phy

alejoe91:
https://github.com/SpikeInterface/spiketoolkit/issues/306

### DIFF
--- a/spiketoolkit/postprocessing/postprocessing_tools.py
+++ b/spiketoolkit/postprocessing/postprocessing_tools.py
@@ -767,7 +767,7 @@ def set_unit_properties_by_max_channel_properties(recording, sorting, property, 
             sorting.set_unit_property(unit_id, property, recording.get_channel_property(max_chan, property))
 
 
-def export_to_phy(recording, sorting, output_folder, n_comp=3, electrode_dimensions=None,
+def export_to_phy(recording, sorting, output_folder, n_comp=3, compute_pc_features=True, max_spikes_per_unit=300,
                   grouping_property=None, ms_before=1., ms_after=2., dtype=None, amp_method='absolute', amp_peak='both',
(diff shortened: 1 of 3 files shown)


spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -902,14 +904,29 @@ def export_to_phy(recording, sorting, output_folder, n_comp=3, electrode_dimensi
         print('Run:\n\nphy template-gui ', str(output_folder / 'params.py'))
 
 
-def _compute_templates_similarity(templates):
+def _compute_templates_similarity(templates, template_ind):
```
samuelgarcia: Note not related to PR:
This is an important function that should not be hidden and private.

spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -1342,6 +1376,27 @@ def _select_max_channels(wf, recording, max_channels):
     return max_channel_idxs
 
 
+def _select_max_channels_from_templates(template, recording, max_channels):
```
samuelgarcia: I don't like the method. Why a max_channels, this is not robust to channel density.
We should either:
  * have a radius
  * a threshold on peak by channel.
Or best both.

spiketoolkit/postprocessing/postprocessing_tools.py
alejoe91: There is another or for that

spiketoolkit/postprocessing/postprocessing_tools.py
alejoe91: That won't work because Phy requires all templates to have the same shape. So we have to set a number of channels

alejoe91: also note that this is fine as the main goal is to find channels for each unit in which we can see differences / similarities. 16 channels are usually enough

PR #320, opened 2020-04-02
Update postprocessing_tools, export_to_phy

llobetv:
Change export_to_phy, removing raw file pathway in params.py.
There is no need and if file is moved raw data is not find anmymore.


colehurwitz:
Looks good! Should we merge @alejoe91 ?

alejoe91:
Yepp thanks @llobetv :)

PR #322, opened 2020-04-05
fix export_to_phy template_ind

llobetv:
Hi :)
If compute_pc_features was False templates_ind.py was wrong. 
Because it used pc_feature_ind to be calculated and that the new method for calculate template ind without pca_ind was wrong (work with template without all channels). 
So I use channel_list return when waveform are exctracted that appear to be the same as pc_feature_ind. 

I don't remove the old method l 1462 but maybe we can?

### DIFF
--- a/spiketoolkit/postprocessing/postprocessing_tools.py
+++ b/spiketoolkit/postprocessing/postprocessing_tools.py
@@ -1459,10 +1459,6 @@ def _get_phy_data(recording, sorting, compute_pc_features, max_channels_per_temp
             templates_ind = pc_feature_ind
         elif channel_list is not None:
             templates_ind = np.array(channel_list)
-        else: # this part doesn't seems to work if template do not cover all channels, remove it?
-            templates_ind = np.zeros((len(templates), max_channels_per_template), dtype='int')
-            for i, temp in enumerate(templates):
-                templates_ind[i] = _select_max_channels_from_templates(temp, recording, max_channels_per_template)
     else:
         templates_ind = np.tile(np.arange(recording.get_num_channels()), (len(sorting.get_unit_ids()), 1))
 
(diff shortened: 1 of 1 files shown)


alejoe91:
HI @llobetv 

I thought I was re-computing template_ind as it is computed in the `get_unit_waveforms`, but it's clearly some duplicate code. I think using `channel_list` as you suggest is better. Can you remove the final else? :)

spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -1457,7 +1457,9 @@ def _get_phy_data(recording, sorting, compute_pc_features, max_channels_per_temp
         # waveforms, templates, and pc_scores are computed on the same channels
         if pc_feature_ind is not None:
             templates_ind = pc_feature_ind
-        else:
+        elif channel_list is not None:
+            templates_ind = np.array(channel_list)
+        else: # this part doesn't seems to work if template do not cover all channels, remove it?
```
alejoe91, line 1462: yep just remove this. Using `channel_list` is more consistent

llobetv:
I make the change and use channel_list first. 
However there is one issue which is if channel_list and pc_feature_ind are None, I don't know if it is possible, but in that case it would probably bug?

spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -1494,29 +1495,7 @@ def _select_max_channels_from_waveforms(wf, recording, max_channels):
         max_channel_idxs = np.arange(recording.get_num_channels())
 
     return max_channel_idxs
-
-
-def _select_max_channels_from_templates(template, recording, max_channels):
```
alejoe91, line 1499: can you leave this? I might need it for other functions :)

llobetv:
I also remove _select_max_channels_from_templates() that seems to be created for this purpose. 
Is it ok?

alejoe91:
That should never happen as `channel_list` is always computed by the `get_unit_waveforms`.

spiketoolkit/postprocessing/postprocessing_tools.py
llobetv: ok

alejoe91:
> I also remove _select_max_channels_from_templates() that seems to be created for this purpose.
> Is it ok?

Can you actually leave it for now? could be useful

llobetv:
Ok, done, I just set templates_ind = np.array(channel_list) every times as it should always exist

alejoe91:
Thanks! merging

PR #330, opened 2020-04-08
export_to_phy add bad_channel_ids argument

llobetv:
Hi,

The function export_to_phy don't work if you provide recording without bad channels, so I add it as argument.
The issue with bad channels is that it could be take as best_channel when it's not, only because there is aberrant high value. 
So I template of bad channels to 0 when best channel is calculated.

Did the way I do it is ok for you?

### DIFF
--- a/spiketoolkit/postprocessing/postprocessing_tools.py
+++ b/spiketoolkit/postprocessing/postprocessing_tools.py
@@ -71,7 +71,7 @@ def update_all_param_dicts_with_kwargs(kwargs):
 
 
 def get_unit_waveforms(recording, sorting, unit_ids=None, channel_ids=None,
-                       return_idxs=False, **kwargs):
+                       return_idxs=False, bad_channel_ids=[], **kwargs):
     '''
     Computes the spike waveforms from a recording and sorting extractor.
 
@@ -87,6 +87,8 @@ def get_unit_waveforms(recording, sorting, unit_ids=None, channel_ids=None,
         List of channels ids to compute waveforms from
     return_idxs: bool
         If True, spike indexes and channel indexes are returned
+    bad_channel_ids: list
+        List of channel id which are bad, prevent from detecting them as best_channel.
     **kwargs: Keyword arguments
         A dictionary with default values can be retrieved with:
         st.postprocessing.get_waveforms_params():
@@ -134,7 +136,7 @@ def get_unit_waveforms(recording, sorting, unit_ids=None, channel_ids=None,
         unit_ids = sorting.get_unit_ids()
     elif not isinstance(unit_ids, (list, np.ndarray)):
         raise Exception("unit_ids is not a valid in valid")
-
+        
     assert np.all([u in sorting.get_unit_ids() for u in unit_ids]), "Invalid unit_ids"
 
     params_dict = update_all_param_dicts_with_kwargs(kwargs)
@@ -171,7 +173,7 @@ def get_unit_waveforms(recording, sorting, unit_ids=None, channel_ids=None,
                 n_channels = _get_max_channels_per_waveforms(recording, grouping_property, channel_ids,
                                                              max_channels_per_waveforms)
                 max_channel_idxs = _select_max_channels_from_waveforms(waveforms, recording,
-                                                                       n_channels)
+                                                                       n_channels, bad_channel_ids)
                 spike_index_list.append(indexes)
                 channel_index_list.append(max_channel_idxs)
     else:
@@ -224,7 +226,8 @@ def get_unit_waveforms(recording, sorting, unit_ids=None, channel_ids=None,
                                                                                        max_channels_per_waveforms,
                                                                                        max_spikes_per_unit, n_pad,
                                                                                        dtype, seed, verbose,
-                                                                                       memmap_array=arr)
+                                                                                       memmap_array=arr, 
+                                                                                       bad_channel_ids=bad_channel_ids)
                     waveform_list.append(arr)
                     spike_index_list.append(indexes)
                     channel_index_list.append(max_channel_idxs)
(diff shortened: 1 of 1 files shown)


alejoe91:
HI @llobetv 

There's a nice preprocessor that does exactly what you are suggesting:

`rec_rm_bad = st.preprocessing.remove_bad_channels(rec, bad_channel_ids=[0, 2, 5])`

This outputs a new wxtractor with bad channels removed. That should be exactly what you need right?

Cheers,
Alessio

alejoe91:
I recommend removing bad channels before any preprocessing/sorting :) you can also use an automatic approach which removes channels with too high std with respect to average std

llobetv:
I try with st.preprocessing.remove_bad_channels(rec, bad_channel_ids=[0, 2, 5]) but it doesn't work with export_to_phy function because index of the bad channel doesn't exist anymore.

llobetv:
Exctract waveform don't return anything for the bad channel, when the best thing for export_to_phy would be trace with 0 everywhere. It would prevent bug and

alejoe91:
Ok I'll look into that! I think the best thing would be to remove the channel completely. I used it before but it might be broken with the latest updates

alejoe91:
Hi @llobetv 

I could not reproduce the issue with master. I tried this script and it works fine on my side:
```
import spikeextractors as se
import spiketoolkit as st

rec, sort = se.example_datasets.toy_example(num_channels=32)
rec_rm = st.preprocessing.remove_bad_channels(rec, bad_channel_ids=[1,3,5])
print(rec_rm.get_num_channels())
st.postprocessing.export_to_phy(rec_rm, sort, 'phy_test', max_channels_per_template=6)
```

Launching phy correctly shows some missing channels, but templates are extracted correctly:

image

Honestly I don't see how it should fail. The extractor returned by the `remove_bad_channels` doesn't know about the missing channels. It just know that it has (in this case) 29 channels. So I don't understand your argument about missing indexes. Now the indexes instead of being 0-31 (for 32 channels), are 0-28 (for 29 channels)

alejoe91:
I think it's also better to remove the channel rather than displaying all zeros. Inf act the straight zero line could be hard to distinguish from templates that are very low (e.g. on channels far away). 
What error do you get by the way? can you print the error trace?

llobetv:
This code :
```
import spikeextractors as se
import spiketoolkit as st

dead_channels = [22]

recording = se.BinDatRecordingExtractor(file_path, sampling_frequency, numchan, dtype)
recording = st.preprocessing.remove_bad_channels(recording, bad_channel_ids)
recording = recording.load_probe_file(prb_file)

recording = si.load_binary_recording(path + "data.dat", sampling_frequency, 
                                     numchan, dtype, 
                                     prb_file=path+"64_para.prb",
                                     bad_channel_ids=dead_channels)

st.postprocessing.export_to_phy(recording, sorting, 
                                outuput_path, n_comp=2, 
                                compute_pc_features=False,
                                grouping_property=None, 
                                ms_before=5, ms_after=5, dtype=None, 
                                amp_method='absolute', amp_peak='both', 
                                amp_frames_before=3, amp_frames_after=3, 
                                max_spikes_per_unit=100, 
                                max_spikes_for_pca=10000, 
                                max_channels_per_template=16, 
                                recompute_info=False, save_features_props=True, 
                                verbose=True, memmap=True, seed=0, n_jobs=1) 

```
Give me this error:
```

Waveform 55/196
Waveform 56/196
Waveform 57/196
Waveform 58/196
Traceback (most recent call last):

  File "/home/llobet/Documents/python/NeurAlgo-ExtracellularAnalysis-master/spikeSorting/spikeinterface_class.py", line 309, in <module>
    bad_channel_ids=[])

  File "/home/llobet/Documents/python/NeurAlgo-ExtracellularAnalysis-master/spikeSorting/spikeinterface_class.py", line 264, in export_to_phy
    bad_channel_ids=bad_channel_ids)

  File "/home/llobet/miniconda3/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 991, in export_to_phy
    max_channels_per_template, bad_channel_ids=bad_channel_ids, **kwargs)

  File "/home/llobet/miniconda3/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 1421, in _get_phy_data
    seed=seed, compute_pc_features=compute_pc_features)

  File "/home/llobet/miniconda3/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 1240, in _get_quality_metric_data
    bad_channel_ids=bad_channel_ids)

  File "/home/llobet/miniconda3/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 229, in get_unit_waveforms
    bad_channel_ids=bad_channel_ids)

  File "/home/llobet/miniconda3/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 1706, in _extract_waveforms_one_unit
    max_channel_idxs = _select_max_channels_from_waveforms(wf, recording, max_channels_per_waveforms, bad_channel_ids)

  File "/home/llobet/miniconda3/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 1492, in _select_max_channels_from_waveforms
    loc_max = locs[recording.get_channel_ids().index(max_channel_idx)]

ValueError: 22 is not in list
```

I used my modified code but hte error was the same before

llobetv:
max_channel_idx is calculated based on a template without a channel but call an index (which don't exist).

llobetv:
In your case channel idx seems to be rename after bad channels were removed?

alejoe91:
What is the output of `recording.get_channel_ids()`? 

I think I found the problem! I don't think it's the bad channels. Instead [this line](https://github.com/SpikeInterface/spiketoolkit/blob/master/spiketoolkit/postprocessing/postprocessing_tools.py#L1480) should be:

```
loc_max = locs[max_channel_idx]
```

instead of:
```
loc_max = locs[recording.get_channel_ids().index(max_channel_idx)]
```

Since `max_channel_idx` is an index, not a channel id!!! Makes sense?

alejoe91:
I pushed the fix here: https://github.com/SpikeInterface/spiketoolkit/pull/332

Can you try to see if that works?

Sorry about that!

llobetv:
It fix the bug!

It was the first solution I tried but i made a mistake that make me thought it was more complicated to solve...

Thanks :)

alejoe91:
I'll close this then! :)

PR #333, opened 2020-04-14
(WIP) Compute waveform features 

alejoe91:
@TRuikes 

Here is a skeleton :)

PR #334, opened 2020-04-14
Waveform features

TRuikes:
Hi! 
I have made a draft to [compute through to peak duration](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L144-L194) and I would like your feedback to make sure I'll be implementing it according to your standards.

A few thoughts:
* the low level functions (computing 1 metric) should require sorting to have the 'template' property. As you mentioned this morning, I'm not expecting people to do it for waveforms. In the future you could expand the module to take any array/list, but for now developping will be easier for me if I assume it's allready in sorting.

* to deal with the requirement 'needs template', I added the function [add_template](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L93-L108), I haven't checked if something like this allready exists

* the low level functions dont take `**kwargs` for readability

* the goal of this module is mainly to have the [compute_unit_template_features](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L16-L25) available for the users, which should handle all top-down inputs

My intended workflow is displayed in [this notebook](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/workfile.ipynb), let me know if it's aligned with spikeinterface.

Then two direct questions:
1. How do you want to deal with undefined values ([like when the peak is detected before the through](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L179))

2. Do we know the 'main channel of detection' per unit? If so, how do I access it?


alejoe91:
> Hi!
> I have made a draft to [compute through to peak duration](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L144-L194) and I would like your feedback to make sure I'll be implementing it according to your standards.
> 

Sounds good!

> A few thoughts:
> 
> * the low level functions (computing 1 metric) should require sorting to have the 'template' property. As you mentioned this morning, I'm not expecting people to do it for waveforms. In the future you could expand the module to take any array/list, but for now developping will be easier for me if I assume it's allready in sorting.

I think we can do it this way. You can add an arg `recording=None`: if the sorting already has `template` properties, then you can just pull those. If not, you can call the `st.postprocessing.get_unit_templates(recording, sorting)`, asserting that `recording` is not None. 
The `get_unit_templates()` will automatically save templates as properties, so the acual computation will be done once (same for waveforms).
Basically, I would start every low-level function by checking if the `template` property exist, and if it doesn't calling the `get_unit_templates()`
What do you think? 


> * to deal with the requirement 'needs template', I added the function [add_template](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L93-L108), I haven't checked if something like this allready exists

Given the previous point, I don't think that the `add_template` is needed as the `get_unit_templates()` does exactly the same

> * the low level functions dont take `**kwargs` for readability

I think it's ok to add **kwargs (but I can add it later, don't worry about that)

> * the goal of this module is mainly to have the [compute_unit_template_features](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L16-L25) available for the users, which should handle all top-down inputs
> 
> My intended workflow is displayed in [this notebook](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/workfile.ipynb), let me know if it's aligned with spikeinterface.
> 

Looks great! Let's discuss about that further. I think it's easier with a call

> Then two direct questions:
> 
> 1. How do you want to deal with undefined values ([like when the peak is detected before the through](https://github.com/TRuikes/spiketoolkit/blob/2293b0f8091355841b6aa2d5e1675dd02f532f4e/spiketoolkit/postprocessing/features.py#L179))

We can discuss edge cases. What I would do is look for positive peaks *following* the negative peak on the max channel.


> 2. Do we know the 'main channel of detection' per unit? If so, how do I access it?

You can get the max_channel for each unit with the [`get_unit_max_channels()`](https://github.com/SpikeInterface/spiketoolkit/blob/master/spiketoolkit/postprocessing/postprocessing_tools.py#L399) function :)

spiketoolkit/postprocessing/features.py
```diff
@@ -96,149 +115,202 @@ def compute_unit_template_features(waveforms,
...
+#                    timestamps[peak_idx]
+#     else:
+#         duration = timestamps[trough_idx:][np.where(waveform[trough_idx:] == np.max(waveform[trough_idx:]))[0][0]] - \
+#                    timestamps[trough_idx]
+#
+#     return duration * 1e3
 
-    return repolarization_slope * 1e-6
 
+def compute_trough_to_peak_duration(sorting,
```
samuelgarcia, line 144: is "peak_to_valey" a more common naming, no ?

spiketoolkit/postprocessing/features.py
```diff
@@ -96,149 +115,202 @@ def compute_unit_template_features(waveforms,
...
 
-    peak_idx = np.argmax(waveform[max_point:]) + max_point
+    for uid in unit_ids:
+        templates = sorting.get_unit_property(uid, 'template')
+        unit_durations = []
+        unit_peaks = []
+        unit_throughs = []
 
-    recovery_slope = linregress(timestamps[peak_idx:peak_idx + window], waveform[peak_idx:peak_idx + window])[0]
+        for i, channel_template in enumerate(templates):
```
alejoe91, line 169: You can do this without the loop with:
`trough_idx  = np.argmin(templates, axis=1)`

spiketoolkit/postprocessing/features.py
```diff
@@ -96,149 +115,202 @@ def compute_unit_template_features(waveforms,
...
-# ==========================================================
+        durations.append(unit_durations)
+        peaks.append(unit_peaks)
+        throughs.append(unit_throughs)
 
-# EXTRACTING 2D FEATURES
+    if save_property_or_features:
+        for i, uid in enumerate(unit_ids):
+            sorting.set_unit_property(uid, 'trough_to_peak_duration', durations[i])
+            sorting.set_unit_property(uid, 'peak_index_in_template', peaks[i])
```
alejoe91, line 188: I would just save the `through_to_peak_duration` (the indexes are trivial and fast to recompute)

spiketoolkit/postprocessing/plot_features.py
```diff
@@ -0,0 +1,163 @@
+from plotly import graph_objs as go
```
alejoe91, line 1: THis is super cool, but I think it should go in [`spikewidgets`](https://github.com/SpikeInterface/spikewidgets)

alejoe91: (for now we are using matplotlib as backend in SW)

TRuikes:
Thanks for the feedback, all is clear for me. One question I formulated wrong:
* How do you want to deal with undefined values (like when the peak is detected before the through),
where I mean, what would you like the function to output if the metric is undefined (np.nan, None).

In the case of TTP; do you prefer hard coding to detect the peak after through? My take was to say; if peak before trough, it's undefined (which can be valid for channels further away from the main channel. Then we can set boundaries per metric as to what is valid. Then the meaning of the metric is _is valid and has value_

The alternative is that we force the requirements (ie peak_has_ to be after through), in which case the output should be considered _is probably valid and has value_.

Let's not plan a meeting just yet, but keep end of the week in the backs of our head.

spiketoolkit/postprocessing/plot_features.py
TRuikes: I don't intend this to become part of `spiketoolkit` it's now there for my convenience, but I was planning on using this for my own analysis

spiketoolkit/postprocessing/features.py
TRuikes: I agree that's it's cheap to compute, but I was thinking maybe you might want to visualize in the future what your computations have done, at which point it's neater to not have the computation at a second place in your repository.

But what should be where, and how it's named, I'll leave up to you guys. For development I would like to keep it here, but once this module is finished we can easily strip these.

spiketoolkit/postprocessing/features.py
TRuikes: to be honest, I don't know =). I'm fine with any, I'll follow your suggestions

alejoe91: I think both are used...we can discuss about naming later :)

spiketoolkit/postprocessing/features.py
alejoe91: 👍

alejoe91:
> Thanks for the feedback, all is clear for me. One question I formulated wrong:
> 
> * How do you want to deal with undefined values (like when the peak is detected before the through),
>   where I mean, what would you like the function to output if the metric is undefined (np.nan, None).
> 

I'd rather use np.nan

> In the case of TTP; do you prefer hard coding to detect the peak after through? My take was to say; if peak before trough, it's undefined (which can be valid for channels further away from the main channel. Then we can set boundaries per metric as to what is valid. Then the meaning of the metric is _is valid and has value_
> 
> The alternative is that we force the requirements (ie peak_has_ to be after through), in which case the output should be considered _is probably valid and has value_.
> 

I see your point, and maybe it's worth considering adding some args (e.g. `positive_after_negative_only`).

> Let's not plan a meeting just yet, but keep end of the week in the backs of our head.

Sounds good. Just drop me an email when you want to meet!

TRuikes:
I have more or less finished what I had in mind. I made [waveform_features_library](https://github.com/TRuikes/spiketoolkit/blob/waveform_features/spiketoolkit/postprocessing/waveform_features_library.py) , by a lack of better name, which contains the actual computation on the lowest level. Each function here takes a 2D waveform array and a bit of meta information. I think this way, we can on a user level directly point to this script if they want to know how metrics were computed, rather than them having to plow trough a series of nested modules/classes. Another pro is that this gives us space to do input/output handling in [compute_waveform_features](https://github.com/TRuikes/spiketoolkit/blob/waveform_features/spiketoolkit/postprocessing/compute_waveform_features.py).

Example function call [in this notebook](https://github.com/TRuikes/spiketoolkit/blob/waveform_features/workfile.ipynb). I haven't verified the metrics yet, I'll be working on that the next days.

I wanted to impose some constraints on the input data, using [_get_trough_and_peak_idx](https://github.com/TRuikes/spiketoolkit/blob/5354e711ab8716342ea870ec0cba162b55402ea5/spiketoolkit/postprocessing/waveform_features_library.py#L96-L111). Further the whole library assumes troughs to be negative, we can control input either in the library using a [_check_wave_form](https://github.com/TRuikes/spiketoolkit/blob/5354e711ab8716342ea870ec0cba162b55402ea5/spiketoolkit/postprocessing/waveform_features_library.py#L114) or one level higher.

Finally, what are you standard output units ([s]/[ms], [V]/[mV])?

alejoe91:
HI @TRuikes 

I was just checking it out and it looks good.

> I have more or less finished what I had in mind. I made [waveform_features_library](https://github.com/TRuikes/spiketoolkit/blob/waveform_features/spiketoolkit/postprocessing/waveform_features_library.py) , by a lack of better name, which contains the actual computation on the lowest level. Each function here takes a 2D waveform array and a bit of meta information. I think this way, we can on a user level directly point to this script if they want to know how metrics were computed, rather than them having to plow trough a series of nested modules/classes. Another pro is that this gives us space to do input/output handling in [compute_waveform_features](https://github.com/TRuikes/spiketoolkit/blob/waveform_features/spiketoolkit/postprocessing/compute_waveform_features.py).
> 

Instead of `waveform_features_library` I would call it `waveform_features_utils`. I would also expand the `compute_waveform_features` to allow for computation of single features, but I can add it later.

I think that we could allow to upsample the template by n factor. I think this would result in better estimates for durations, halfwidth, and even peak values.

> Example function call [in this notebook](https://github.com/TRuikes/spiketoolkit/blob/waveform_features/workfile.ipynb). I haven't verified the metrics yet, I'll be working on that the next days.
> 
> I wanted to impose some constraints on the input data, using [_get_trough_and_peak_idx](https://github.com/TRuikes/spiketoolkit/blob/5354e711ab8716342ea870ec0cba162b55402ea5/spiketoolkit/postprocessing/waveform_features_library.py#L96-L111). Further the whole library assumes troughs to be negative, we can control input either in the library using a [_check_wave_form](https://github.com/TRuikes/spiketoolkit/blob/5354e711ab8716342ea870ec0cba162b55402ea5/spiketoolkit/postprocessing/waveform_features_library.py#L114) or one level higher.
> 

I think it's fair to assume that waveforms are negative for the moment. We have tools that allow to invert the signals in case that is not the case. I would skip the `check_waveform` for now. Let's keep some checks on the trough peak.

> Finally, what are you standard output units ([s]/[ms], [V]/[mV])?

Sampling frequency is in Hz, but I would use ms by default. We don't make explicit assumptions on the singals (even though most of them are in uV). The functions should just return the same unit as the input template

I'll write other suggestions as comments, but it's a great start!

TRuikes:
Allright, do the comments appear on this page? or in the code? I'll start checking the output friday.

spiketoolkit/postprocessing/waveform_features_library.py
```diff
@@ -0,0 +1,253 @@
+"""
+Functions based on
+https://github.com/AllenInstitute/ecephys_spike_sorting/blob/master/ecephys_spike_sorting/modules/mean_waveforms/waveform_metrics.py
+15/04/2020
+"""
```
alejoe91, line 5: Example comment @TRuikes: 

1. We could allow to upsample the templates
2. And to compute features on all channels (or a subset of maximum channels)

TRuikes: 1. how would you interpolate the signal?
2. `waveform` in all functions can be a 2D matrix, I'd handle what exact channels to pass one lever higher

TRuikes:
basecode is now in SpikeInterface/spikefeatures, continue on #333

PR #335, opened 2020-04-15
Removing epochs from quality metrics [WIP]

colehurwitz:
WIP on removing epochs from quality metrics --> replacing them with manual subsorting and subrecording extractors.

@alejoe91 I ran into a little issue with the `waveforms_idxs`  spike feature... for some reason it is stored as a list instead of a numpy array so when you call get_epoch on a sorting extractor with this feature, it will throw an error. I can't find where this feature is set. Could you point me to the relevant code?

### DIFF
--- a/spiketoolkit/tests/test_validation.py
+++ b/spiketoolkit/tests/test_validation.py
@@ -2,41 +2,41 @@ import spikeextractors as se
 import numpy as np
 from spiketoolkit.validation import compute_isolation_distances, compute_isi_violations, compute_snrs, \
     compute_amplitude_cutoffs, compute_d_primes, compute_drift_metrics, compute_firing_rates, compute_l_ratios, \
-    compute_metrics, compute_nn_metrics, compute_num_spikes, compute_presence_ratios, compute_silhouette_scores, \
+    compute_quality_metrics, compute_nn_metrics, compute_num_spikes, compute_presence_ratios, compute_silhouette_scores, \
     get_validation_params
 
 
 def test_functions():
     rec, sort = se.example_datasets.toy_example(duration=10, num_channels=4, seed=0)
 
-    firing_rates = compute_firing_rates(sort, seed=0)[0]
-    num_spikes = compute_num_spikes(sort, seed=0)[0]
-    isi = compute_isi_violations(sort, seed=0)[0]
-    presence = compute_presence_ratios(sort, seed=0)[0]
-    amp_cutoff = compute_amplitude_cutoffs(sort, rec, seed=0)[0]
-    max_drift, cum_drift = compute_drift_metrics(sort, rec, seed=0, memmap=False)[0]
-    silh = compute_silhouette_scores(sort, rec, seed=0)[0]
-    iso = compute_isolation_distances(sort, rec, seed=0)[0]
-    l_ratio = compute_l_ratios(sort, rec, seed=0)[0]
-    dprime = compute_d_primes(sort, rec, seed=0)[0]
-    nn_hit, nn_miss = compute_nn_metrics(sort, rec, seed=0)[0]
-    snr = compute_snrs(sort, rec, seed=0)[0]
-    metrics = compute_metrics(sort, rec, return_dict=True, seed=0)
+    firing_rates = compute_firing_rates(sort, rec.get_num_frames(), seed=0)
+    num_spikes = compute_num_spikes(sort, rec.get_num_frames(), seed=0)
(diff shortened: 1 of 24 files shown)


alejoe91:
@colehurwitz what's the status of this?

colehurwitz:
Pretty much done! Will hopefully finish today

alejoe91:
👍 

Merge with master before as I pushed some changes!

colehurwitz:
This requires the spikemetrics PR to work.

spiketoolkit/validation/quality_metrics.py
```diff
@@ -22,6 +22,7 @@
 
 def compute_num_spikes(
         sorting,
+        duration,
```
alejoe91, line 25: This is not used. I think MetricData could allow duration=None (like in this case)

colehurwitz: Good point!

colehurwitz:
Ok, I merged spikemetrics and will get this to pass now

spiketoolkit/validation/quality_metrics.py
alejoe91: Maybe we can also allow duration to be `None`. If not given max(spiketimes) - min(spiketimes) will be used. What do you think?

colehurwitz: Oh, I thought we wanted duration to be mandatory. If we allow for it to be optional, then the value will change if you pass a RecordingExtractor in or not.

alejoe91: maybe `duration` suggests the duration in seconds. `duration_in_frames`? 
I think we could also add `get_duration()` in the `RecordingExtractor` (maybe in the `SortingExtractor` too, similarly to sampling_frequency.

colehurwitz: I can call it `num_frames`. We already have `get_num_frames` in the RX which is equivalent since all of our extractors start at frame 0.

colehurwitz:
Ready to merge @alejoe91

spiketoolkit/validation/quality_metrics.py
colehurwitz: Called it `duration_in_frames` as you mentioned.

alejoe91:
can you release spikemetrics first and increment the requirement?

colehurwitz:
> can you release spikemetrics first and increment the requirement?

Sure

colehurwitz:
tagged and pushed version 0.1.5 of spikemetrics

colehurwitz:
Sorry, what requirement do you want me to increment? Spiketoolkit doesn't depend on a specific version of spikemetrics, right?

alejoe91:
I think we should force the requirement because the API has changed

colehurwitz:
Ok, I can force the requirement to be 0.1.5 and higher

colehurwitz:
Done! Ready to merge

colehurwitz:
I can update documentation afterwards in spikeinterface

alejoe91:
Did you change the min_isi?

alejoe91:
Let me take a closer look tomorrow morning an I'll merge :) you can start updating the docs!

colehurwitz:
I was waiting on changing the min_isi until we resolved the discussion about refractory window times in the open issue. I think that should be a separate PR where we fix all isi stuff.

spiketoolkit/tests/test_validation.py
```diff
@@ -2,41 +2,41 @@
...
+    presence = compute_presence_ratios(sort, rec.get_num_frames(), seed=0)
+    amp_cutoff = compute_amplitude_cutoffs(sort, rec, seed=0)
+    max_drift, cum_drift = compute_drift_metrics(sort, rec, seed=0, memmap=False)
+    silh = compute_silhouette_scores(sort, rec, seed=0)
+    iso = compute_isolation_distances(sort, rec, seed=0)
+    l_ratio = compute_l_ratios(sort, rec, seed=0)
+    dprime = compute_d_primes(sort, rec, seed=0)
+    nn_hit, nn_miss = compute_nn_metrics(sort, rec, seed=0)
+    snr = compute_snrs(sort, rec, seed=0)
+    metrics = compute_quality_metrics(sort, rec, return_dict=True, seed=0)
```
alejoe91, line 24: `return_dict` should be removed, but I can make a separate PR :)

PR #343, opened 2020-04-22
Fixes to compute similarity and recompute_info for phy

alejoe91:
- `_compute_template_similarity` did nt consider the swap axis
- when `recompute_info` is True, also clear `pca_scores` features
- add warning message is recording is not filtered

@llobetv @Dradeliomecus you probably have a wrong similarity score in phy due to this! I suggest re-exporting to phy :)


llobetv:
Hi @alejoe91, 

we are still in the programming step (make tools to improve our analysis) so their is not so much to recompute. 

Thank you very much for thinking of us :)
Cheer,
Victor

PR #346, opened 2020-04-23
(WIP) spikefeatures

TRuikes:
Hi! 
These are the TODO's I think:
-update docstring
-decide on loop structure (right now it loops over all units, computed features for all channels, then only saves the main features)
-polish output format (align with metrics)
-update kwargs
-make methods to compute single metrics

### DIFF
--- a/spiketoolkit/postprocessing/features.py
+++ b/spiketoolkit/postprocessing/features.py
@@ -11,7 +11,7 @@ from scipy.stats import linregress
 from scipy.signal import resample
 
 
-def compute_waveform_metrics(waveforms,
+def compute_unit_template_features(waveforms,
                                cluster_id,
                                peak_channel,
                                channel_map,
@@ -67,12 +67,12 @@ def compute_waveform_metrics(waveforms,
 
     timestamps = np.linspace(0, num_samples / sample_rate, new_sample_count)
 
-    duration = compute_waveform_duration(mean_1D_waveform, timestamps)
-    halfwidth = compute_waveform_halfwidth(mean_1D_waveform, timestamps)
-    PT_ratio = compute_waveform_PT_ratio(mean_1D_waveform)
-    repolarization_slope = compute_waveform_repolarization_slope(
+    duration = compute_unit_template_duration(mean_1D_waveform, timestamps)
+    halfwidth = compute_unit_template_halfwidth(mean_1D_waveform, timestamps)
+    PT_ratio = compute_unit_template_PT_ratio(mean_1D_waveform)
+    repolarization_slope = compute_unit_template_repolarization_slope(
         mean_1D_waveform, timestamps)
-    recovery_slope = compute_waveform_recovery_slope(
+    recovery_slope = compute_unit_template_recovery_slope(
         mean_1D_waveform, timestamps)
 
     amplitude, spread, velocity_above, velocity_below = compute_2D_features(
@@ -96,7 +96,7 @@ def compute_waveform_metrics(waveforms,
 # ==========================================================
 
 
-def compute_waveform_duration(waveform, timestamps):
+def compute_unit_template_duration(waveform, timestamps):
     """
     Duration (in seconds) between peak and trough
     Inputs:
@@ -122,7 +122,7 @@ def compute_waveform_duration(waveform, timestamps):
     return duration * 1e3
 
 
-def compute_waveform_halfwidth(waveform, timestamps):
+def compute_unit_template_halfwidths(waveform, timestamps):
     """
     Spike width (in seconds) at half max amplitude
     Inputs:
@@ -160,7 +160,7 @@ def compute_waveform_halfwidth(waveform, timestamps):
     return halfwidth * 1e3
 
 
-def compute_waveform_PT_ratio(waveform):
+def compute_unit_template_pt_ratios(waveform):
     """
     Peak-to-trough ratio of 1D waveform
     Inputs:
@@ -180,7 +180,7 @@ def compute_waveform_PT_ratio(waveform):
     return PT_ratio
 
 
-def compute_waveform_repolarization_slope(waveform, timestamps, window=20):
+def compute_unit_template_repolarization_slopes(waveform, timestamps, window=20):
     """
     Spike repolarization slope (after maximum deflection point)
     Inputs:
@@ -204,7 +204,7 @@ def compute_waveform_repolarization_slope(waveform, timestamps, window=20):
     return repolarization_slope * 1e-6
 
 
-def compute_waveform_recovery_slope(waveform, timestamps, window=20):
+def compute_unit_template_recovery_slopes(waveform, timestamps, window=20):
     """
     Spike recovery slope (after repolarization)
     Inputs:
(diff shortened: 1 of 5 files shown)


alejoe91:
Hi @TRuikes 

> -update docstring

Yup, but last thing to do

>-decide on loop structure (right now it loops over all units, computed features for all channels, 

I think we can select units and channel_ids and just run the fnctions on those. This is done already for units, and it can be added for channel_ids as well (no need to compute all if one asks for one channel!)

> then only saves the main features)

OK

> -polish output format (align with metrics)

OK

> -update kwargs

I can do that

> -make methods to compute single metrics

I can do that


I would add:

- [ ] Add tests

The proper tests should be implemented in `spikefeatures`. Here we just make sure that all functions run with no errors. I can take care of it.

spiketoolkit/postprocessing/features.py
```diff
@@ -0,0 +1,143 @@
...
+        factor with which to upsample the template resolution (default 1)
+
+    TODO:
+    save_property_or_features=False,
+    recovery_slope_window=0.7,
+    invert_waveforms=False,
+    spread_threshold=None,
+    site_range=None,
+    site_spacing=None,
+    epoch_name=None,
```
alejoe91, line 68: epochs have been finally removed from quality metrics

spiketoolkit/postprocessing/features.py
```diff
@@ -0,0 +1,143 @@
...
+        list of feature names to be computed
+    upsampling_factor : int
+        factor with which to upsample the template resolution (default 1)
+
+    TODO:
+    save_property_or_features=False,
+    recovery_slope_window=0.7,
+    invert_waveforms=False,
+    spread_threshold=None,
+    site_range=None,
```
alejoe91, line 66: what are these for? (site range and site spacing).
If for 2d features, they are already available through the recording extractor

spiketoolkit/postprocessing/features.py
```diff
@@ -0,0 +1,143 @@
...
+            columns: metrics
+
+    OR
+    TODO
+    """
+
+    # ------------------- SETUP ------------------------------
+    if unit_ids is None:
+        unit_ids = sorting.get_unit_ids()
+
```
alejoe91, line 87: Same thing for `channel_ids` here and then loop only through them

spiketoolkit/postprocessing/features.py
```diff
@@ -0,0 +1,143 @@
...
+    TODO:
+    save_property_or_features=False,
+    recovery_slope_window=0.7,
+    invert_waveforms=False,
+    spread_threshold=None,
+    site_range=None,
+    site_spacing=None,
+    epoch_name=None,
+    channel_map=None,
```
alejoe91, line 69: what is channel map for?

spiketoolkit/postprocessing/features.py
TRuikes: i dont use them, ill remove

spiketoolkit/postprocessing/features.py
TRuikes: removed

spiketoolkit/postprocessing/features.py
TRuikes: ok, but should we call the function using `unit_ids` OR `channel_ids`?

spiketoolkit/postprocessing/features.py
TRuikes: not sure, something with channel select, ill rename to `channel_ids`

spiketoolkit/postprocessing/features.py
TRuikes: and now also from features!

spiketoolkit/postprocessing/features.py
TRuikes: How does the `channel_ids` input argument look like? Can we have multiple channels per unit?

spiketoolkit/postprocessing/features.py
TRuikes: yea left them for 2d features, removed now

spiketoolkit/postprocessing/features.py
TRuikes: nothing yet, we'll continue with `channel_ids`

TRuikes:
I just realized that the spiketrain statistics also depend on the time interval you compute them on. So it might be a bit too specific to implement here. Maybe better to leave them out here, and have people use elephant themselves for Fano Factor and Local Variance

spiketoolkit/postprocessing/features.py
TRuikes: I think we can do either:
pass `unit_ids` and `channel_ids`; `channel_ids` is a `dict`  or a `list with lists` with id's per channel

or 

Only allow 1 channel per unit; `channel_ids` is a list of length `unit_ids`

In either case, if  `channel_ids` are not given it defaults to the detected max_channel

alejoe91: No, it's a single list. Otherwise in `get_unit_waveforms` we also have a `max_channels_per_waveforms` arg that selects the max channels separately for each unit.

Don't worry about that for now, I can add it later (and I think it coud be useful)

TRuikes: or to ask this differently; do you see use for people passing multiple channels per unit?

TRuikes: allright

alejoe91: I do, but I think it would be easier to select the best channels internally (already implemented [here](https://github.com/SpikeInterface/spiketoolkit/blob/master/spiketoolkit/postprocessing/postprocessing_tools.py#L1657-L1676)).

So in my opinion one could:
- pass `channel_ids`, i.e. a single list of channels used for all unit_ids
- do not pass `channel_ids` and pass `max_channels_per_features`, which selects the `max_channels_per_features` best channels for that unit 
- I think that by default, `max_channel_per_features` should be 1, as most people probably extract feats from the max channel

What do you think?

alejoe91: If `channel_ids` is used, it overrides `max_channel_per_waveforms`

alejoe91:
@TRuikes I can take it from here if you want :)

TRuikes:
Yes perfect! I have made changes according to above, and for me it's finished now!

alejoe91:
> Yes perfect! I have made changes according to above, and for me it's finished now!

Thanks again! I'll make a new PR soon with final fixes and tests and will let you know :)

spiketoolkit/postprocessing/features.py
```diff
@@ -0,0 +1,170 @@
...
+        unit_ids = sorting.get_unit_ids()
+
+    if channel_ids is None:
+        channel_ids = get_unit_max_channels(
+            sorting=sorting,
+            recording=recording,
+            unit_ids=unit_ids,
+        )
+    else:
+        assert len(channel_ids) == len(unit_ids), f'#channel ids should be # unit ids'
```
alejoe91, line 95: this is wrong. Channel ids independent of unit ids. Imagine you have 100 channels and 10 units: you can select any subset of the 100 channels

alejoe91: I'll fix it later :)

alejoe91:
Can you merge agin with master before I merge? just to doubel check

TRuikes:
git pull https://github.com/SpikeInterface/spiketoolkit waveform_features gives me an 'allready up to date' 
is that the correct way to check?

TRuikes:
that worked =)

PR #348, opened 2020-04-24
subset of units improved to reduce computation

colehurwitz:
Computation of metrics performed only on specified units. Identical values as doing all the units, but should be more efficient!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
         'numpy',
         'spikeextractors',
         'spikesorters',
-        'spikemetrics >= 0.1.6',
+        'spikemetrics >= 0.2.0',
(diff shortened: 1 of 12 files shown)


colehurwitz:
Needs spikemetrics PR to work

setup.py
```diff
@@ -22,7 +22,7 @@
         'numpy',
         'spikeextractors',
         'spikesorters',
-        'spikemetrics >= 0.1.5',
+        'spikemetrics >= 0.1.6',
```
alejoe91: isn't it 0.2.0?

colehurwitz: Updated in latest commits

colehurwitz:
merging

Issue #355, opened 2020-04-28
How to delete spikes in CurationSortingExtractor?


Dradeliomecus:
Hi,

I am trying to remove certain spikes (similar to what I was doing in #339).
The solution of using the `CacheSortingExtractor` worked really well, and now I am trying to do the same but with the `CurationSortingExtractor`.
I successfully removed the spikes, but when doing the `export_to_phy()` method, I have the following error:
```
Traceback (most recent call last):
  File "remove_duplicated_spikes.py", line 24, in <module>
    st.postprocessing.export_to_phy(recording, sorting, output_folder=save_folder, verbose=True, n_jobs=1, ms_before=3, ms_after=4, compute_pc_features=False, max_spikes_per_unit=2000, recompute_info=False, memmap=True)
  File "/home/aurelien/dev/miniconda3/envs/phy/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 1088, in export_to_phy
  File "/home/aurelien/dev/miniconda3/envs/phy/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 1517, in _get_phy_data
  File "/home/aurelien/dev/miniconda3/envs/phy/lib/python3.7/site-packages/spiketoolkit-0.6.1-py3.7.egg/spiketoolkit/postprocessing/postprocessing_tools.py", line 1429, in _get_quality_metric_data
ValueError: could not broadcast input array from shape (234166,1) into shape (234092,1)
```

From what I could gather, I think it has to do with the problem that I deleted the spikes but didn't update the amplitude array as well, which lead to a mismatch?

Here is how I deleted the spikes:
```
def del_spikes_from_unit(sorting, spikes, unit_id):
    if not isinstance(sorting, st.curation.CurationSortingExtractor):
        print("Converting sorting...")
        sorting = st.curation.CurationSortingExtractor(sorting)

    # Need to find the Unit in "roots" that matches our unit_id
    unit = "error"
    for root in sorting._roots:
        if root.unit_id == unit_id:
            unit = root
            break

    if unit == "error": # Didn't find the Unit with our unit_id
        print("Error in del_spikes_from_unit():")
        print("Unit ID {0} not found".format(unit_id))
        return

    spike_train = unit.get_spike_train()
[...]

PR #356, opened 2020-04-28
return unit ids curation sorting extractor

colehurwitz:
#355 

Split and merge in the `CurationSortingExtractor` now return the new unit_ids as requested.


colehurwitz:
@alejoe91

PR #357, opened 2020-04-29
Update curationsorting.py

llobetv:
Hi,
I Implemented idxs method for features in split_unit()

Cheers,
Victor

### DIFF
--- a/spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
+++ b/spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
@@ -236,9 +236,34 @@ class CurationSortingExtractor(SortingExtractor):
             self._roots.append(new_root_2)
 
             for feature_name in self.get_unit_spike_feature_names(unit_id):
+                if feature_name.endswith('_idxs'):
+                    continue
                 full_features = np.array(self.get_unit_spike_features(unit_id, feature_name))
-                self.set_unit_spike_features(new_root_1_id, feature_name, full_features[indices_1])
-                self.set_unit_spike_features(new_root_2_id, feature_name, full_features[indices_2])
+                if not feature_name + '_idxs' in self.get_unit_spike_feature_names(unit_id):
+                    self.set_unit_spike_features(new_root_1_id, feature_name, full_features[indices_1])
+                    self.set_unit_spike_features(new_root_2_id, feature_name, full_features[indices_2])
+                else:
+                    full_features_idxs = np.array(self.get_unit_spike_features(unit_id, feature_name+'_idxs'))
+                    indices_1_idxs = [n for n, i in enumerate(full_features_idxs) if i in indices]
+                    indices_2_idxs = [n for n, i in enumerate(full_features_idxs) if not i in indices]
+                    
+                    # Calc new idxs after split
+                    indexes_1 = []
+                    indexes_2 = []
+                    for i in full_features_idxs:
+                        if i in indices:
+                            indexe_1 = np.count_nonzero(np.array(indices)<i)
+                            indexes_1.append(indexe_1)
+                        else:
+                            indexe_2 = i - np.count_nonzero(np.array(indices)<i)
+                            indexes_2.append(indexe_2)                            
+                            
+                    self.set_unit_spike_features(new_root_1_id, feature_name, 
+                                                 full_features[indices_1_idxs],
+                                                 indexes=indexes_1)
+                    self.set_unit_spike_features(new_root_2_id, feature_name, 
+                                                 full_features[indices_2_idxs],
+                                                 indexes=indexes_2)
             del self._features[unit_id]
             del self._roots[root_index]
         else:
@@ -290,3 +315,4 @@ class Unit(object):
         for child in self.children:
             ret += child.__str__(level + 1)
         return ret
+
--- a/spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
+++ b/spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
@@ -246,8 +246,8 @@ class CurationSortingExtractor(SortingExtractor):
                     self.set_unit_spike_features(new_root_2_id, feature_name, full_features[indices_2])
                 else:
                     full_features_idxs = np.array(self.get_unit_spike_features(unit_id, feature_name+'_idxs'))
-                    indices_1_idxs = [n for n, i in enumerate(full_features_idxs) if i in indices]
-                    indices_2_idxs = [n for n, i in enumerate(full_features_idxs) if not i in indices]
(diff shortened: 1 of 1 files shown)


llobetv:
For testing:
```
sorting_path = ""
recording_path = ""
sorting = se.PhySortingExtractor(sorting_path)
recording = se.PhyRecordingExtractor(recording_path)
# This line exctract subpart of waveforms, which then will be save with idxs in sorting obj
waveforms = \
    st.postprocessing.get_unit_waveforms(recording, sorting,
                                         ms_before=1e3,
                                         ms_after=1e3, 
                                         max_channels_per_waveforms=1, 
                                         max_spikes_per_unit=10,
                                         memmap=True,
                                         n_jobs=1,
                                         verbose=False,
                                         save_property_or_features=True,
                                         recompute_info=True)

child_sorting = st.validation.quality_metric_classes.utils.curationsortingextractor.CurationSortingExtractor(sorting)
indices = range(int(len(sorting.sorting.get_unit_spike_train(0))/2)) # split in 2
child_sorting.split_units(0, indices)  
```

spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
```diff
@@ -236,9 +236,34 @@ def split_unit(self, unit_id, indices):
             self._roots.append(new_root_2)
 
             for feature_name in self.get_unit_spike_feature_names(unit_id):
+                if feature_name.endswith('_idxs'):
+                    continue
                 full_features = np.array(self.get_unit_spike_features(unit_id, feature_name))
```
colehurwitz, line 241: I wouldn't cast this as a numpy array since we don't want to load it all into memory.

colehurwitz:
Looks good! One change that we are adding to indices is that they will be sorted now in ascending order. Not sure if that would help with your split units function.

spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
llobetv: It was already cast as np.array(), I'll check if how can do it without that. 
Their is the same thing in merge function, I'll check for it too.

alejoe91: you can check the type: if it's list, then you can cast it as np.array, otherwise you leave it as it is

llobetv: it's already numpy array ^^, so i remove np.array()

llobetv: And in fact their is nothing to change in merge funciton, my bad

alejoe91: you can remove the np.array from the full_features and add:
```
if isinstance(full_features, list):
    full_features = np.array(full_features)
```
That should make sure all the indexing works properly

llobetv: Thanks :p
Done

spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
```diff
@@ -236,9 +236,35 @@ def split_unit(self, unit_id, indices):
             self._roots.append(new_root_2)
 
             for feature_name in self.get_unit_spike_feature_names(unit_id):
-                full_features = np.array(self.get_unit_spike_features(unit_id, feature_name))
-                self.set_unit_spike_features(new_root_1_id, feature_name, full_features[indices_1])
-                self.set_unit_spike_features(new_root_2_id, feature_name, full_features[indices_2])
+                if feature_name.endswith('_idxs'):
+                    continue
+                if isinstance(full_features, list):
+                    full_features = np.array(full_features)
```
alejoe91, line 242: this fails as `full_features` is not defined.
You should add this line before:
`full_features = self.get_unit_spike_features(unit_id, feature_name)`

alejoe91: @llobetv

llobetv: I was too fast sorry, fix :)

spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
```diff
@@ -236,9 +236,36 @@ def split_unit(self, unit_id, indices):
...
+                    continue
+                full_features = self.get_unit_spike_features(unit_id, feature_name)
+                if isinstance(full_features, list):
+                    full_features = np.array(full_features)    
+                if not feature_name + '_idxs' in self.get_unit_spike_feature_names(unit_id):
+                    self.set_unit_spike_features(new_root_1_id, feature_name, full_features[indices_1])
+                    self.set_unit_spike_features(new_root_2_id, feature_name, full_features[indices_2])
+                else:
+                    full_features_idxs = np.array(self.get_unit_spike_features(unit_id, feature_name+'_idxs'))
+                    indices_1_idxs = [n for n, i in enumerate(full_features_idxs) if i in indices]
```
alejoe91, line 249: these should be parsed as np.array otherwise using them as idxs will fail in line 263:

`indices_1_idxs = [n for n, i in enumerate(full_features_idxs) if i in indices]`
`indices_2_idxs = [n for n, i in enumerate(full_features_idxs) if i not in indices]`

spiketoolkit/validation/quality_metric_classes/utils/curationsortingextractor.py
```diff
@@ -236,9 +236,36 @@ def split_unit(self, unit_id, indices):
...
+                    indexes_1 = []
+                    indexes_2 = []
+                    for i in full_features_idxs:
+                        if i in indices:
+                            indexe_1 = np.count_nonzero(np.array(indices)<i)
+                            indexes_1.append(indexe_1)
+                        else:
+                            indexe_2 = i - np.count_nonzero(np.array(indices)<i)
+                            indexes_2.append(indexe_2)                            
+
```
alejoe91, line 262: same here: it's safer to always parse indexes to arrays:
```
indexes_1 = np.array(indexes_1)
indexes_2 = np.array(indexes_2)
```

llobetv:
Fix, 
But I don't really understand why their is a bug if index is a list ?
I didn't encounter any issue with that when I launch my code.

Cheers,
Victor

alejoe91:
Still failing, maybe you should make sure the indexes are of type int. You can test with pytest locally if tests pass :)

llobetv:
Found, it was because full_features was a range (not a list so not converted to np.array)
I add dtype=int for all indices :p

PR #359, opened 2020-05-04
small speed-up getting waveforms

mhhennig:
Sort spike indices before requesting snippets is slightly faster (esp. hdf5).


alejoe91:
@mhhennig can you apply the same change here:  https://github.com/SpikeInterface/spiketoolkit/blob/master/spiketoolkit/postprocessing/postprocessing_tools.py#L553 ?

samuelgarcia:
As I said somewhere a major speedup would be to grab snippet grouped by index proximity and not by unid_id. Because by index imply to read the same buffer for all units.

Issue #360, opened 2020-05-04
Some praticals issues with dead channels and similarity matrix


llobetv:
Hi,

I encounter some praticals issues.

Fisrt link to dead channels. If I set a dead channels, it won't be exctract which is nice but when phy show data it will generated some painful behavior. The channel exctract instead of the dead one will change for every channel and in my case switch from right to left every channel (and make waveform representation move a lot every cluster, which make it paintful to use).

So I just decided to exctract all channel, but then similarity score just become really bad because it use proportion of shared exctracted channel for ponderation. It would be much better to use real distance ?

Cheers,
Victor

alejoe91:
Hi Victor,

Sorry I didn't get exactly what's the problem here. Can you provide an example/a phy plot?
What painful behavior do you observe in phy? The problem with phy is that all templates must have the same shape, so an extra channel must be chosen (usually we pick the closest one to the largest peak).

llobetv:
Hi Alejoe,

I know the pb, but then the channel will switch form right to left each time I switch channel (which happen often) wich is a bit anoying for intensive analysis (that will begin soon). 
ex:

![image](https://user-images.githubusercontent.com/22896067/81546229-5b94e800-937a-11ea-8777-0d5df454d84c.png)
![image](https://user-images.githubusercontent.com/22896067/81546264-664f7d00-937a-11ea-828b-9d0de2ce4a98.png)

So of course exctract all channel solve this issue but then similarity score totally failes because is based on n channel exctract. 
I can try an implementation to add an adjency_radius to scale similarity value.

Cheer,
Victor

alejoe91:
@llobetv I have an idea. Instead of appending an extra channel, we can just append an empty template of nans. This should keep the. shape (so phy is happy) and we can easily disable it when computing similarity

llobetv:
Hi,
The solution you propose would work well I think, and solve all the issue I mention before !

alejoe91:
Hi @llobetv 

Do you have time to give it a try? I think that the main problem is when you ave multiple groups. Is it your case?

llobetv:
I can give a try but not before Tuesday,

Yes it is, I actually have 4 groups of 16 channels.

alejoe91:
[...]

PR #365, opened 2020-05-20
Update postprocessing.py

llobetv:
Correction for #360 
Change template exctraction for grouping_property, now if some template are missing to reach the right number of channel, it's fill with 0, and channel idx with -1.
Change compute_template_similarity to avoid using these channels.

### DIFF
--- a/spiketoolkit/postprocessing/postprocessing_tools.py
+++ b/spiketoolkit/postprocessing/postprocessing_tools.py
@@ -1131,9 +1131,10 @@ def _compute_templates_similarity(templates, template_ind=None):
     similarity = np.zeros((len(templates), len(templates)))
     if template_ind is None:
         template_ind = np.tile(np.arange(templates[0].shape[0]), (len(templates), 1))
+    
     for i, (t_i, t_ind_i) in enumerate(zip(templates, template_ind)):
         for j, (t_j, t_ind_j) in enumerate(zip(templates, template_ind)):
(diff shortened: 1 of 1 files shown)


llobetv:
I try with NaN but for teamplate phy crash, and for channel idx, it was change to abherant negative value somewhere in the code.

llobetv:
Here to test:

```
import pytest
import spikeextractors as se
from spiketoolkit.postprocessing import export_to_phy
import os
import shutil
from spiketoolkit.preprocessing import remove_bad_channels

@pytest.mark.implemented
def test_export_to_phy():
    folder = 'test'
    if os.path.isdir(folder):
        shutil.rmtree(folder)
    rec, sort = se.example_datasets.toy_example(dump_folder=folder, dumpable=True, duration=10, num_channels=8)
    rec.set_channel_groups([0, 0, 0, 0, 1, 1, 1, 1])
    rec = remove_bad_channels(rec, [1,2, 5])    
    export_to_phy(rec, sort, output_folder='phy_group', grouping_property='group', recompute_info=True)



if __name__ == '__main__':
    # test_waveforms()
    test_export_to_phy()
```

llobetv:
An exemple:
image
image

alejoe91:
@llobetv testing it now and it looks good :)

alejoe91:
It sounds a bit strange that with -1 it works, because -1 is the last index, but maybe this is built-in in phy to discard channels eith -1

PR #369, opened 2020-05-28
Fix num spikes validation adn curation is empty units

alejoe91:
@colehurwitz threshold_num_spikes was failing if empty units. This fixes it


colehurwitz:
This works! It took me a second to notice what the issue was, but I see what you were trying to solve. Maybe worth calling it 'fail_if_empty'?

PR #388, opened 2020-08-31
Test on windows

alejoe91:


setup.py
```diff
@@ -21,7 +21,6 @@
     install_requires=[
         'numpy',
         'spikeextractors',
-        'spikesorters',
```
samuelgarcia, line 24: Oh yes.
You catch it!
This was totally wird before when installing from sourcs : the spikesorter package was was installed form pypi anyway.

alejoe91: that was not supposed to be there. spiketoolkit doesn't use spikesorters.

Can you check why the tests are failing? I guess it's because the str should be a raw str...

samuelgarcia: my guess is that we try to open an existing file not already close with w+.

samuelgarcia: Or a pathlib stuff that should str.

alejoe91: > Or a pathlib stuff that should str.

impossible: https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/baseextractor.py#L222

I think the path needs to be `r'path-to-file'` for windows, even if pathlib takes care of `\\`

alejoe91:
This might fix the tests: https://github.com/SpikeInterface/spikeextractors/pull/450

samuelgarcia:
Can we merge this ?
The 'spikesorters' in sorter is meesing up my installation ?

PR #392, opened 2020-09-24
fix setup.py

samuelgarcia:


samuelgarcia:
@alejoe91 I need to fix this soon because it force to install the pypi version of spikesorters evryttime and I want my local version.

PR #399, opened 2020-10-14
Speed up waveforms extraction

alejoe91:
Refractor of `get_waveforms` based on @magland 's implementation in [labbox-ephys](https://github.com/laboratorybox/labbox-ephys/blob/6f0d24c4eaef52c448833b570476ee632143db7c/python/labbox_ephys/helpers/get_unit_waveforms.py#L84)

In brief, traces are extracted in chunks and loaded into memory, all waveforms are extracted and then concatenated. From the original implementation, I added the possibility to select a subset of spikes, to deal with grouping properties and max_channels_for_waveforms, and to memmap (default).

The speed increase is a lot. On a 1-min Neuropixels with 254 units:

- old implementation 1 job: ~ 5482s
- old implementation 8 jobs: ~ 1643s
- **new implementation 1 job: ~ 83s**


spiketoolkit/postprocessing/utils.py
```diff
@@ -124,3 +124,78 @@ def get_max_channels_per_waveforms(recording, grouping_property, channel_ids, ma
...
+    if (0 <= start_frame) and (end_frame <= traces.shape[1]):
+        x = traces[:, start_frame:end_frame]
+    else:
+        # handle edge cases
+        x = np.zeros((traces.shape[0], end_frame - start_frame), dtype=traces.dtype)
+        i1 = int(max(0, start_frame))
+        i2 = int(min(traces.shape[1], end_frame))
+        x[:, (i1 - start_frame):(i2 - start_frame)] = traces[:, i1:i2]
+    if channel_indices is not None:
+        x = x[channel_indices, :]
```
samuelgarcia, line 144: Use `x = np.take(x, indices, axis=0)` you should have a small speedup.

samuelgarcia:
Well done camarade.
There is no more parrallalisation ?

spiketoolkit/postprocessing/utils.py
alejoe91: @samuelgarcia instead of what?

samuelgarcia: x = x[channel_indices, :] is sugar syntax for x = np.take(x, indices, axis=0)
but calling directly is a slightly bit faster
if you do many many time (the case) here you could use the function (and maybe keep your line commented)

alejoe91:
@samuelgarcia @mhhennig 

So I added parallelization in this way: given the `chunk_mb` limit, the number of chunks is increased `n_jobs` times to use the same amount of RAM. Then the different chunks are processed in parallel. In case of memmap, the jobs write directly to a shared memmap file.

However, the speed does not seem to decrease so much for small number of spikes (300), probably due to the overhead time of spawning the jobs.

- Benchmark 1) Neuropixels (~280 chanels), 15 min duration, 258 units, ALL SPIKES, chunk_mb - 500Mb
    - 1 job: ~ 20min
    - 4 jobs: ~ 10min
In both cases, the max RAM usage is around 4Gb, due to the loading of waveforms for each chunk in memory before dumping to the memmap file. Note that in this case the disk space used by the memmap objects is around 100 Gb!!!

- Benchmark 2) Neuropixels (~280 chanels), 15 min duration, 258 units, 300 SPIKES, chunk_mb - 500Mb
    - 1 job: ~ 183s
    - 4 jobs: ~ 153s
Max RAM usage < 1Gb

alejoe91:
@oliche 

This PR makes waveform extraction (and in turn also metric computation) MUCH faster.
Could you test it in one of your Neuropixels datasets?

spiketoolkit/postprocessing/utils.py
alejoe91: @samuelgarcia that has been removed. Waveforms are extracted on all channels and trimmed later in case it's needed. Thanks anyways for the suggestion :)

alejoe91:
IMO this is ready to be merged!!!

samuelgarcia:
Alessio.
I would add some more info in the docstring of get_unit_waveforms to explain the strategy of this snippet extraction.
ordering spike/looping over memory block.
So that the reading for further modification would be easier.

samuelgarcia:
cool

PR #420, opened 2020-11-27
Return ec feature as scalar if max_channels=1

alejoe91:


alejoe91:
This is in line with how metrics are handled.

PR #430, opened 2020-12-14
fix #428

### LINKED ISSUE #428: BlankSaturationRecording for memmaps
Hi guys, I notice that inside BlankSaturationRecording the values obtained by self._recording.get_traces() are edited. And that coulb be a memmap. I try using it with a recording that I loaded using load_extractor_from_pickle() and I got:

>    traces[traces >= self._threshold] = self._median
> ValueError: assignment destination is read-only

I guess that a solution could be: check the type of traces and if it is a memmap and it will be edited then a copy on ram should be made.

ferchaure:
this is just the change we talk  about on #428

### DIFF
--- a/spiketoolkit/preprocessing/blank_saturation.py
+++ b/spiketoolkit/preprocessing/blank_saturation.py
@@ -46,6 +46,8 @@ class BlankSaturationRecording(BasePreprocessorRecordingExtractor):
         traces = self._recording.get_traces(channel_ids=channel_ids,
                                             start_frame=start_frame,
                                             end_frame=end_frame)
+        if isinstance(traces, np.memmap):
+            traces = traces.copy()
         if self._lower:
             traces[traces <= self._threshold] = self._median
         else:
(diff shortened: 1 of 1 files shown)


samuelgarcia:
I think we have to copy every time because it modify the source!!!!
Even in memory object it is very dangerous because if we use the recording for other stuff then it is in place modify.
Quite bad.
I propose a `copy=True` by **default** in all situations. If if the user really want a inplace modification then we can do `copy=False`

ferchaure:
you are probably right. are you saying an argument `copy ` on get_traces method or in the BlankSaturationRecording constructor?

alejoe91:
In my opinion we should use copy all the times. When one gets traces after a processing step the it's expected that the traces are loaded into RAM. I wouldn't add an extra copy argument. Do you guys agree?

ferchaure:
I really hate to use extra ram but always copy is the only secure option. I will make the change

alejoe91:
Sorry @ferchaure I forgot to merge this. Merged now! Thanks again :)

PR #435, opened 2021-01-01
Improve noise overlap metric computation

khl02007:
@alejoe91: I tried to make the computing of noise overlap metric faster by (1) computing the noise snippets just once (rather than computing it separately for every unit) by moving it outside the main for-loop; and (2) using signal only at a specified number of channels when projecting to PC space (similar to the other PC-based metrics like nearest neighbor). This required giving a `num_channels_to_compare` parameter to `NoiseOverlap.compute_metrics` and I made the relevant changes to handle this. Let me know how this looks.


alejoe91:
@khl02007 the problem as suspected came from an update of scipy that broke the `decimate` function. Since we have been discussing about removing the `decimate` function in favor of `resample`, I implemented it here. Should fix tests

spiketoolkit/validation/quality_metric_classes/noise_overlap.py
```diff
@@ -23,47 +26,75 @@ def __init__(self, metric_data):
         if not metric_data.has_recording():
             raise ValueError("MetricData object must have a recording")
 
-    def compute_metric(self, max_spikes_per_unit_for_noise_overlap, num_features, num_knn, **kwargs):
+    def compute_metric(self, num_channels_to_compare, max_spikes_per_unit_for_noise_overlap,
+                        num_features, num_knn, **kwargs):
+
+        # update keyword arg in case it's already specified to something
+        kwargs['max_spikes_per_unit'] = max_spikes_per_unit_for_noise_overlap
+        kwargs['recompute_info'] = True
```
alejoe91, line 34: Why is this always set to True?
```suggestion
```

spiketoolkit/validation/quality_metric_classes/noise_overlap.py
```diff
@@ -23,47 +26,75 @@ def __init__(self, metric_data):
...
+        # waveforms = List (units,) of np.array (n_spikes, n_channels, n_timepoints)
         waveforms = st.postprocessing.get_unit_waveforms(
             self._metric_data._recording,
             self._metric_data._sorting,
             unit_ids=self._metric_data._unit_ids,
-            max_spikes_per_unit=max_spikes_per_unit_for_noise_overlap,
             **kwargs
         )
 
+        # set random seed
```
alejoe91, line 48: if waveforms are already computed, then retrieving them is very quick. The problem is (and I think this is why you initially set `recompute_info=True`) that they could have been computed on less than `max_spikes_per_unit_for_noise_overlap`. In this case, I would do something like this:

```
n_waveforms_per_unit = [len(wf) for wf in waveforms]
n_spikes_per_unit = [len(self._metric_data._sorting.get_unit_spike_train(u)) for u in self._metric_data._unit_ids]

if np.all(n_waveforms_per_unit < max_spikes_per_unit_for_noise_overlap) and 
    np.any(n_spikes_per_unit > max_spikes_per_unit_for_noise_overlap):
    # in this case it means that waveforms have been computed on 
    # less spikes than max_spikes_per_unit_for_noise_overlap --> recompute
    kwargs['recompute_info'] = True
    waveforms = st.postprocessing.get_unit_waveforms(
            self._metric_data._recording,
            self._metric_data._sorting,
            unit_ids=self._metric_data._unit_ids,
            max_spikes_per_unit=max_spikes_per_unit_for_noise_overlap,
            **kwargs
        )
elif np.all(n_waveforms_per_unit > max_spikes_per_unit_for_noise_overlap):
    # waveforms computed on more spikes than needed --> sample
    for i_w, wfs in enumerate(waveforms):
        if len(wfs) > max_spikes_per_unit_for_noise_overlap:
                selecte_idxs = np.random.choice(times, size=max_spikes_per_unit_for_noise_overlap)
                waveforms[i] = wfs[selecte_idxs]
```

This makes sure that you always have `max_spikes_per_unit_for_noise_overlap` spikes without recomputing waveforms (if not needed).

spiketoolkit/validation/quality_metric_classes/noise_overlap.py
```diff
@@ -23,47 +26,75 @@ def __init__(self, metric_data):
         if not metric_data.has_recording():
             raise ValueError("MetricData object must have a recording")
 
-    def compute_metric(self, max_spikes_per_unit_for_noise_overlap, num_features, num_knn, **kwargs):
+    def compute_metric(self, num_channels_to_compare, max_spikes_per_unit_for_noise_overlap,
+                        num_features, num_knn, **kwargs):
+
```
alejoe91, line 31: Since this implementation relies on `max_spikes_per_unit_for_noise_overlap` not being `None`, I would add:
```
assert max_spikes_per_unit_for_noise_overlap is not None, "'max_spikes_per_unit_for_noise_overlap' must be an integer."
```

spiketoolkit/validation/quality_metric_classes/noise_overlap.py
```diff
@@ -23,47 +26,75 @@ def __init__(self, metric_data):
...
+        if num_channels_to_compare > len(channel_idx):
+            num_channels_to_compare = len(channel_idx)
+
+        # get noise snippets
+        min_time = min([self._metric_data._sorting.get_unit_spike_train(unit_id=unit)[0]
+                    for unit in self._metric_data._sorting.get_unit_ids()])
+        max_time = max([self._metric_data._sorting.get_unit_spike_train(unit_id=unit)[-1]
+                    for unit in self._metric_data._sorting.get_unit_ids()])
+        times_control = np.random.choice(np.arange(min_time, max_time),
```
alejoe91, line 64: What if every unit has less spikes than `max_spikes_per_unit_for_noise_overlap`? Safe to add:

```
max_spikes = np.max([len(self._metric_data._sorting.get_unit_spike_train(u)) for u in self._metric_data._unit_ids])

if max_spikes < max_spikes_per_unit_for_noise_overlap:
    max_spikes_per_unit_for_noise_overlap = max_spikes
```

alejoe91:
@khl02007 thank you very much about the PR! I have added a few comments

spiketoolkit/validation/quality_metric_classes/noise_overlap.py
khl02007: Added this

spiketoolkit/validation/quality_metric_classes/noise_overlap.py
khl02007: Added this

spiketoolkit/validation/quality_metric_classes/noise_overlap.py
khl02007: Yes that's precisely why I added `recompute_info=True`. Added your suggestion.

khl02007:
Thanks for the feedback, I incorporated your suggestions and updated the PR.

khl02007:
@alejoe91  Actually I had to make a change in the if statement for computing the waveforms because the test wouldn't pass.

alejoe91:
@khl02007 thanks! Can you fix the test? Seems to be a problem with `default_rng `: https://travis-ci.org/github/SpikeInterface/spiketoolkit/jobs/755271344

khl02007:
> @khl02007 thanks! Can you fix the test? Seems to be a problem with `default_rng `: https://travis-ci.org/github/SpikeInterface/spiketoolkit/jobs/755271344

Done

PR #440, opened 2021-01-16
Add return_ scaled argument to preprocessors

alejoe91:


samuelgarcia:
Alessio.
This approah is strange.
The question should be : "what do we expect from a traces non scale than we filetered" ?
And not : "how do we retrieve the input data scaled or not ?"

alejoe91:
Sam this is purely because the return_scaled is now in the Base rec, and we need to add the arg here too otherwise the decorator throws an error.

samuelgarcia:
Yes but I think here as do are doing preprocessing we should do something like assert return_scaled=True, 'None scaled are not supported in preprocessing functions' and internally we should make a call to the input with return_scaled=False

alejoe91:
Why do you think it should be False by default?

samuelgarcia:
No True and if False would raise an error.

alejoe91:
> No True and if False would raise an error.

I don't think it should throw an error. It's perfectly fine to filter int16...

Issue #448, opened 2021-02-02
spiketoolkit.preprocessing.remove_artifacts.RemoveArtifactsRecording traces can't be used


joramvanrheede:
After I have used st.preprocessing.remove_artifacts() on a recordingextractor, I am unable to use the get_traces() method on the new 'RemoveArtifactsRecording' extractor or use spikewidgets.plot_timeseries() to visualise my artifact-free traces.

In both cases I get with the following error:

> File "... my_env\lib\site-packages\spiketoolkit\preprocessing\remove_artifacts.py", line 28, in get_traces
>     traces[:, trig - pad[0]:trig + pad[1]] = 0
> ValueError: assignment destination is read-only

I assume it has somethig to do with the fact that the RemoveArtifactsRecording extractor uses a memmap rather than having the whole trace available, but is there a way to get around this?

alejoe91:
Hi @joramvanrheede 

We had the same problem with another preprocessor and ended up calling a `copy()` of the traces before applying the changes. Will make a PR tomorrow, unless you want to give it a try?

joramvanrheede:
Hi Alessio,

Thanks for the quick reply - I'm happy to give it a try tomorrow afternoon! 

Joram

alejoe91:
https://github.com/SpikeInterface/spiketoolkit/pull/449

PR #449, opened 2021-02-03
RemoveArtifacts: Copy traces to fix write access

joramvanrheede:
Make a copy of memmapped traces in memory to be able to access it for editing when the traces are requested using the get_traces() method

Fixes issue #448 - ValueError: assignment destination is read-only


alejoe91:
Thanks @joramvanrheede :)

PR #452, opened 2021-02-14
Simple fast high pass filter

mhhennig:
Runs fast as simple Butterworth order 1 filter, use it to remove slow, LFP-like fluctutations (also has fft version). Faster than the bandpass. Note that generally the fft based method seems slower (it's default in the bandpass).


alejoe91:
@mhhennig can you provide some benchmarks in terms of speed of the different approaches?

Issue #451, opened 2021-02-04
Replacing artifacts with interpolation


joramvanrheede:
I'm currently dealing with a data set that involves electrical stimulation, and the artifacts are HUGE. They also seriously affect the quality of the spike sorting.

Currently remove_artifacts() sets parts of the traces affected by artifacts to 0. Therefore, running it on the raw signal (which often fluctuates some distance away from 0) produces a trace that has a lot of sudden deviations to 0 that are problematic in their own right.

Running remove_artifacts() on a high-pass filtered recording (obtained using bandpass_filter() ) gets rid of this issue, but when artifacts are very large (as in my current recordings) there is filter ringing behaviour that affects a wider part of the trace. 

For my purposes it would therefore be very helpful to replace artifacts in the raw trace using some kind of interpolation between the start and the end of the artifact window.

I'm going to have a go at making my own 'interpolate_artifacts.py' with an 'InterpolateArtifactsRecording' class - I can generate a pull request if you think this would be of interest to others. 

It looks like the get_traces() method of the RemoveArtifactsRecording class is the crucial place for implementing this - are there any other places you think I should make edits to accomplish this?

alejoe91:
Hi @joramvanrheede 

I think it's a good idea, even though I'm not sure what the best interpolation would be.
I think that the best but not trivial option would be to fit some kind of AR model to the data and predict the missing values using that. 

But sure give it a go with a PR and we can discuss on that! You could also modify the `remove_artifact` by adding an argument (e.g. `mode='zeros' / 'interpolate'`).

Yes, everything you need to modify is the `get_traces()` method (in case you don't need other computations -- e.g. in the `CenterRecording` the center value for each channel is computed in the `__init__`

joramvanrheede:
Thanks Alessio - I will go with your suggestion of adding a 'mode' argument to remove_artifact for now - that does seem more sensible, will report back once I've got something working

joramvanrheede:
Hi @alejoe91 

I've currently got something working using scipy.interpolate.interp1d. 

Using spaced out samples from 1ms before and after the artifact window, I fit a cubic spline and then I replace the artifact samples by the values for that fit at the sample points. It seems to achieve a relatively smooth continuation of the trace. 

(I use spaced out samples because if using only the last 2-3 samples on either side of the artifact window the fit is very susceptible to random fluctuations at that point)

I've accounted for edge cases with no samples before/after by simply filling the nearest available value.

[...]

PR #453, opened 2021-02-17
Add artifact removal modes (linear & cubic interpolation)

joramvanrheede:
Add 'mode' input argument to remove_artifacts(). Default is 'zeros' (old behaviour). If 'linear', implements linear interpolation between pre and post artifact points. If 'cubic' - implements cubit spline interpolation.
Additional input 'fit_sample_spacing' determines how far apart sample points are taken for creating the spline fit that is used to fill in the gap (default is 1 ms). Shorter sample intervals lead to better continuation locally, but may lead to steeper fit slopes with more high frequency content.

Related to issue #451

### DIFF
--- a/spiketoolkit/preprocessing/remove_artifacts.py
+++ b/spiketoolkit/preprocessing/remove_artifacts.py
@@ -133,7 +133,11 @@ class RemoveArtifactsRecording(BasePreprocessorRecordingExtractor):
 
 def remove_artifacts(recording, triggers, ms_before=0.5, ms_after=3, mode = 'zeros', fit_sample_spacing = 1):
     '''
-    Removes stimulation artifacts from recording extractor traces. Artifact periods are zeroed-out.
+    Removes stimulation artifacts from recording extractor traces. By default, 
+    artifact periods are zeroed-out (mode = 'zeros'). This is only recommended 
+    for traces that are centered around zero (e.g. through a prior highpass
+    filter); if this is not the case, linear and cubic interpolation modes are
+    also available, controlled by the 'mode' input argument.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


spiketoolkit/preprocessing/remove_artifacts.py
```diff
@@ -22,21 +25,113 @@ def get_traces(self, channel_ids=None, start_frame=None, end_frame=None):
...
+                else:
+                    # No data to interpolate from on either side of gap;
+                    # Fill with zeros
+                    traces[:,gap_idx] = 0
+                    
         return traces
 
 
-def remove_artifacts(recording, triggers, ms_before=0.5, ms_after=3):
+def remove_artifacts(recording, triggers, ms_before=0.5, ms_after=3, mode = 'zeros', fit_sample_spacing = 1):
```
alejoe91, line 134: @joramvanrheede Can you add some explanation of the different modes in the docstrings? also `fit_sample_spacing` should be added there :)

joramvanrheede: Yes, will do that now. I also see a 'Travis CI build failed' notification below but I'm unfamiliar with debugging this sort of issue...

alejoe91: I updated some stuff on the extractors side and this makes the test failing. I'll fix that. Thanks for the additions!

Give me some other few days to properly test this. In the meanwhile, could you add some mock tests here: https://github.com/SpikeInterface/spiketoolkit/blob/master/spiketoolkit/tests/test_preprocessing.py#L197-L215 ?
Basically similar to what's there already but adding the new option

joramvanrheede:
Hi Alessio, I see you've already sorted the tests now - Thank you for that! I had a look last week but wasn't immediately sure what tests would be appropriate (for the zeroed out artifacts it made sense to check everything was zero but i wasn't sure what would be a good equivalent for the cubic and linear fits, without re-creating the entire fit on a local bit of trace in the test).
I also saw you were more efficient with the scipy import and fixed some code style issues - thanks! This is my first time contributing to a public repository that is not just my own so I'm still learning about conventions...

alejoe91:
No worries Joram! Just made a few changes :) thanks for the contribution! Looking forward to more!

PR #460, opened 2021-02-25
Update curationsortingextractor.py

llobetv:
Hi !

There was some issue with set(range(len(original_spike_train))) - set(indices_1)) method. 
In my case sometimes the order of index return change and then spiketrain is no longer in the right order which generate error later....

Best,
Victor


alejoe91:
Thanks @llobetv!

@colehurwitz can you look into this?

colehurwitz:
Yep, will do

colehurwitz:
This change looks good! just double checked and I think we can merge.

alejoe91:
Thanks for checking @colehurwitz

Issue #461, opened 2021-03-05
save_pc_features all_waveforms takes a lot of disk space


Dradeliomecus:
Hi,

While using `export_to_phy`, I found that if `save_pc_features` is set to `True`, then it saves a file `all_waveforms.raw` containing the same information as all the `waveforms_x.raw`.
This duplicates all the waveforms temporarily saved on disk, which in my case writes an additional `62 GB` (for a total of 124 GB), which takes a lot of space and takes a lot of time to write (unless a NVMe SSD is used).

Is there a way to not rewrite the `all_waveforms memmap`? If so, it seems it could be a good improvement.

alejoe91:
Hi @Dradeliomecus 

This is true but it's needed to compute PCA. Do you have a better solution in mind?

samuelgarcia:
We could estimate the pcd projetion (pca.fit) on a subset.
Then we could pca.transform(wf) on the fly without cahce to disk.

Dradeliomecus:
> We could estimate the pcd projetion (pca.fit) on a subset.
> Then we could pca.transform(wf) on the fly without cahce to disk.

I agree,
If the data is this big, the fit can be done on a subset of spikes ; there is no need to use all of the spikes.

alejoe91:
@Dradeliomecus can you try to make a PR for it?

Dradeliomecus:
I can try.
I'll take be a little while to make it and test it, but hopefully I can make it today :)

PR #462, opened 2021-03-05
Added max_spikes_for_pca_fit argument

Dradeliomecus:
Added the ability to specify a maximum number of spikes per unit to fit the pca while exporting to phy.

I set the default number of spikes to 5,000, but you can change it to whatever you want.
I checked using my dataset, and it works!

The PCA looks pretty much the same as the one when I fit on all the waveforms.


Dradeliomecus:
PR made for #461

PR #470, opened 2021-03-18
Issue 390 local car implementation

### LINKED ISSUE #390: Global demuxed CAR (preprocessing step for neuropixels)
Dear, 
Thank you for this amazing tool. 

We are using it with neuropixels. I see that we already have the possibility to preprocess the data with a common average referencing filter, but for neuropixels it needs to be done by group of channels of similar sample time offsets from each ADC group (because of the time division multiplexing). 
Do you know if we can use the group option ([here](https://github.com/SpikeInterface/spiketoolkit/blob/020655ec4e7e1d242ecf300a3b6fbc45c8f5b507/spiketoolkit/preprocessing/common_reference.py#L12)) to do the global demuxed CAR (like in CatGT for example) ? Do you know if it has been already used like this before ? 

Thank you for your help.

MarineChap:
Remarks : 
- I did not implemented a group option as you (@alejoe91) said in the issue it should be something apply on all channels. Seems weird however - should I add one ? 
- I am still unsure that it is logical to have the local_radius option in common_reference as it is an option only for the local filtering. 

close   #390 

Description of the test case added : 
image

Also, a little refactoring/clean-up of this code would be nice. I can do it in other PR if you want. 
For example, I would put this code in an external private method instead of being repeat  3 times. (Tell me if I am overstepping here :) )
```
  for g in self._groups:
                    new_chans = []
                    for chan in g:
                        if chan in self._recording.get_channel_ids():
                            new_chans.append(chan)
                    new_groups.append(new_chans)
```
Also, I am not an expert of pytest (I am more often using unittest). I notice if the test failed there is no clean up done. I am sure there is some way to specify a tear down method which is activated each time a test is finished whether or not it was a fail or a success.

### DIFF
--- a/spiketoolkit/preprocessing/common_reference.py
+++ b/spiketoolkit/preprocessing/common_reference.py
@@ -3,15 +3,20 @@ import numpy as np
 from .basepreprocessorrecording import BasePreprocessorRecordingExtractor
 from spikeextractors.extraction_tools import check_get_traces_args
 
+from ..utils import get_closest_channels
+
 
 class CommonReferenceRecording(BasePreprocessorRecordingExtractor):
     preprocessor_name = 'CommonReference'
 
-    def __init__(self, recording, reference='median', groups=None, ref_channels=None, dtype=None, verbose=False):
+    def __init__(self, recording, reference='median', groups=None, ref_channels=None,
+                                  num_local_channels= 4, local_radius=4,
+                                  dtype=None, verbose=False):
+
         if not isinstance(recording, RecordingExtractor):
             raise ValueError("'recording' must be a RecordingExtractor")
-        if reference != 'median' and reference != 'average' and reference != 'single':
-            raise ValueError("'reference' must be either 'median' or 'average'")
+        if reference not in ['median', 'average', 'single', 'local']:
+            raise ValueError("'reference' must be either 'median', 'average', 'single' or 'local'")
         self._ref = reference
         self._groups = groups
         if self._ref == 'single':
@@ -33,7 +38,8 @@ class CommonReferenceRecording(BasePreprocessorRecordingExtractor):
         self.verbose = verbose
         BasePreprocessorRecordingExtractor.__init__(self, recording)
         self._kwargs = {'recording': recording.make_serialized_dict(), 'reference': reference, 'groups': groups,
-                        'ref_channels': ref_channels, 'dtype': dtype, 'verbose': verbose}
+                        'ref_channels': ref_channels,  'num_local_channels':num_local_channels, 'local_radius': local_radius,
+                        'dtype': dtype, 'verbose': verbose}
 
     @check_get_traces_args
     def get_traces(self, channel_ids=None, start_frame=None, end_frame=None, return_scaled=True):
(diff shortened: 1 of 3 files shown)


alejoe91:
Thanks @MarineChap !

Looks great :) Please go ahead and do some refractoring as well in this PR. All help is very much appreciated!!

For pytest / unittest we can definitely move to unittest and have a proper tearDown, but let's keep it for another PR.

MarineChap:
Hello, 
While working on the PR, I am wondering if the channels_ids option in get_traces work as intended when mixed with the group option of the common reference. It is not cover by the test. 
But if you vstack all your different traces from the different groups and then keep only the channel index corresponding of the general channel id you are probably not keeping the right channel, no ? 


And second point, since I am not overstepping ;), there is a reason for not using the logging library and a verbose option instead ?

alejoe91:
> Hello,
> While working on the PR, I am wondering if the channels_ids option in get_traces work as intended when mixed with the group option of the common reference. It is not cover by the test.
> But if you vstack all your different traces from the different groups and then keep only the channel index corresponding of the general channel id you are probably not keeping the right channel, no ?
> 
> And second point, since I am not overstepping ;), there is a reason for not using the logging library and a verbose option instead ?

I see what you mean: basically `channel_idxs` should be extracted based of the union between the provided groups. I agree that that use case can be wrong! Thanks for the catch.

Regarding logging, there is no reason really. Just want to mention that we are performing a big refractoring of the code base, so in a few months we'll move all the development [here](https://github.com/SpikeInterface/spikeinterface/pull/88). It could be a good opportunity to start using logging.

MarineChap:
Okay I will work on this channel_idxs fix while I am here.

alejoe91:
Looks good @MarineChap ! Let me know when you think it's ready to be merged :)

MarineChap:
I have nothing else to add, it is okay to merge it. When the next release of this package is planned ?

alejoe91:
I can release a version with the new local CAR next week :)

samuelgarcia:
Hi @MarineChap 
I am in a process of a big refactoring here:
https://github.com/SpikeInterface/spikeinterface/pull/88

I will try to propagate your changes in the "new" spikeinetrface soon.
Please have a look on it in following weeks to be sure that your patch is propagated properly.

MarineChap:
Sorry for the style. I tried to look if a style was specified somewhere in the doc but I did not found it. Do you have a description somewhere (for the next time) ? :) Thanks

alejoe91:
No worries! You are right we should add a style description. I just use the default Python style reformatting in Pycharm (so it takesl literally one-click to reformat), which I think it's the PEP8 style :)

MarineChap:
If at some point, you are looking for an automatic way to reformat your code, I am a big fan of precommit + black for this. We are using it in our lab. Each time, someone commit and push, it automatically reformat everything following your style description.

Also, @samuelgarcia okay,  I will keep an eye on it. :)

samuelgarcia:
Hi @MarineChap ,
I am porting your code in the new API.
I think we have a bug here:
```
distances = [np.linalg.norm(l - locs[id]) for l in locs]
```
`locs` is an array an you try to fetch by "id" instead of by "indice".
Most of the time, it is the same except when you have a subrecording.
Maybe I am wrong.

Coudl you check ?

samuelgarcia:
@MarineChap 
Here the PR that contain more or less the same code:
can you check ?
https://github.com/SpikeInterface/spikeinterface/pull/95

I have a question you use average for 'local' would be relevant to use median instead ?

MarineChap:
Ah you are right if the channel id is not the same as the index it will give a wrong result in subrecording. I did not think of that. 
I guess if you replace it by this that should be good.
` distances = [np.linalg.norm(l - locs[recording.get_channel_ids().index(id)]) for l in locs]`
I tested all tests are passing but I did not test with a subrecording. 

For the average or median, I asked what would be the better to Hanna (the one who did the comparison of different CAR performance. I am only the dev in the team - not an expert here) and she said the average. 
Now if you think median is better, go ahead ! I just ask again the same question but to the whole team this time. I can tell you their preferences about it tomorrow.

MarineChap:
@alejoe91 @samuelgarcia  Should I reopen an PR for this fix, push a new commit here or let you do the modif yourself ?

samuelgarcia:
Yes please, push a commit on the same branch and open a new PR. (Si tu as le courage)
I think Alessio is quite busy and I am concentrated on the new API.

PR #474, opened 2021-03-22
fix channel id to find the nearest channels

MarineChap:
fix after discussion with @samuelgarcia  on the previous PR #470


samuelgarcia:
thanks

PR #478, opened 2021-04-12
fix local CAR

MarineChap:
- add assert when the selecting annulus does not contain channels
- fix verbose option bug

### DIFF
--- a/spiketoolkit/preprocessing/common_reference.py
+++ b/spiketoolkit/preprocessing/common_reference.py
@@ -10,7 +10,7 @@ class CommonReferenceRecording(BasePreprocessorRecordingExtractor):
     preprocessor_name = 'CommonReference'
 
     def __init__(self, recording, reference='median', groups=None, ref_channels=None,
-                 local_radius=(2, 8), dtype=None, verbose=False):
+                 local_radius=(30, 55), dtype=None, verbose=False):
 
         if not isinstance(recording, RecordingExtractor):
             raise ValueError("'recording' must be a RecordingExtractor")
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi @MarineChap .
Do you think this is something that have to be propagated here  https://github.com/SpikeInterface/spikeinterface/tree/big_refactoring ?

MarineChap:
Hi @samuelgarcia 
Yes, the assert in case of wrong radius parameters;  not containing any channel references.

MarineChap:
Also, good to know for you and to add somewhere in the doc (?), after test with the neuropixels, the values equivalent to what catgt is using (aka (2,8) electrodes site) for their own local car (which was the main way to do this filtering type with neuropixels before implementating it here) is **(30,55)**.

samuelgarcia:
Do you think you would have time to make a similar PR on the new API ?

MarineChap:
Yes, I can do that this morning.

alejoe91:
Thanks @MarineChap 

Is it good to go for merging?

MarineChap:
Yes. Thanks.

alejoe91:
Actually, before we merge can you set `local_radius` to (30, 55) as you suggested? I think that (2, 8) is unrealistic and it will fail for almost all probes

MarineChap:
yes of course. That makes sense.

alejoe91:
> @samuelgarcia I should change the default value for the new api too ?

Yes please!

alejoe91:
@MarineChap  can you change it in the function as well: https://github.com/SpikeInterface/spiketoolkit/blob/121ce5ddb9f85f16b2c87937b7e82d62fe2d8198/spiketoolkit/preprocessing/common_reference.py#L139?

MarineChap:
Oups sorry! I amended my commit.

PR #489, opened 2021-06-10
Fix/return scaled in get unit waveforms

sbuergers:
I  noticed that while it is possible to specify `return_scaled=False` in `spiketoolkit.postprocessing.get_unit_waveforms()`, the data is actually read as float in `get_unit_waveforms_for_chunk()` as per the default in `get_traces()`: https://github.com/SpikeInterface/spiketoolkit/blob/877d70fe5395e37ea3f47683937eca6221e85cd1/spiketoolkit/postprocessing/utils.py#L153. It is then cast back to int8 (in my case) here: `_extract_waveforms_one_chunk()`https://github.com/SpikeInterface/spiketoolkit/blob/877d70fe5395e37ea3f47683937eca6221e85cd1/spiketoolkit/postprocessing/postprocessing_tools.py#L2026

To avoid this I have propagated the `return_scaled` parameter down to the children of `spiketoolkit.postprocessing.get_unit_waveforms()`. `pytest test_postprocessing` passed locally.

### DIFF
--- a/spiketoolkit/postprocessing/postprocessing_tools.py
+++ b/spiketoolkit/postprocessing/postprocessing_tools.py
@@ -1954,7 +1954,7 @@ def _template_descending_order(recording, templates, templates_ind):
 
 
 def _extract_waveforms_one_chunk(i, rec_arg, chunks, unit_ids, n_pad, times_in_chunk, cumulative_n_spikes,
-                                 waveforms_file, memmap, dtype, verbose):
+                                 waveforms_file, memmap, dtype, verbose, return_scaled=False):
     chunk = chunks[i]
     times_this_chunk = times_in_chunk[i]
     n_spikes = cumulative_n_spikes[i]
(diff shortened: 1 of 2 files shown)


spiketoolkit/postprocessing/postprocessing_tools.py
```diff
@@ -1989,7 +1989,7 @@ def _template_descending_order(recording, templates, templates_ind):
 
 
 def _extract_waveforms_one_chunk(i, rec_arg, chunks, unit_ids, n_pad, times_in_chunk, cumulative_n_spikes,
-                                 waveforms_file, memmap, dtype, verbose):
+                                 waveforms_file, memmap, dtype, verbose, return_scaled=False):
```
alejoe91, line 1992: Can you keep default to True here?

sbuergers: Sure.

